PR #2, opened 2019-02-02
First draft of the schema definition

NellWaliczek:
This PR adds schema files for the mappings explained in the README.md and fixes bugs in the mapping.json files.  Obviously there's still a ton of work left to do to finalize the actual schema, but at least now we have schema files in which to do it.

My next step will be to add the tests and enable Travis.  I just need to figure out the open source contribution license first.


NellWaliczek:
@toji ok, schema and mappings changes pushed.  I still need to update the readme.md to match though

schemas/dataSources.properties.schema.json
```diff
@@ -0,0 +1,47 @@
...
+            "type": "object",
+            "additionalProperties": false,
+            "required": ["gamepadAxesIndex"],
+            "properties": {
+                "gamepadAxesIndex": { 
+                    "allOf": [{ "$ref": "gamepadIndex.schema.json" }]
+                }
+            }
+        }
+    }
```
toji: One thing that I would like to see added to this mapping is a way to flag an axis or button as "empty". The nature of the Gamepad API's `mapping` system is that if a controller lacks an input it should just leave a gap in the array. For example: The Daydream controller, since it lacks a trigger, would have a button[0] that never reports non-zero values under the current Gamepad API proposal. But there's no native way to differentiate between a button that's not present vs. one that's simply never pressed, so it's an area where the schema would have to step in and help.

schemas/dataSources.properties.schema.json
```diff
@@ -0,0 +1,47 @@
...
+                    "default" : "false"
+                }
+            }
+        },
+        "axis": {
+            "type": "object",
+            "additionalProperties": false,
+            "required": ["gamepadAxesIndex"],
+            "properties": {
+                "gamepadAxesIndex": {
```
toji: Is there anything else that you can see being placed in here at some point? Honestly it feels like this should simply be a `gamepadIndex`, rather than an object with a single (kind of verbose) key that points to a `gamepadIndex`.

schemas/dataSources.schema.json
```diff
@@ -0,0 +1,52 @@
...
+                "type": "object",
+                "additionalProperties": false,
+                "required": ["id", "dataSourceType", "left", "right", "down", "up"],
+                "properties":{
+                    "id": { "allOf": [{ "$ref": "dataSources.properties.schema.json#/definitions/id" }] },
+                    "dataSourceType": { "enum": ["dpadSource"] },
+                    "left": { "allOf": [{ "$ref": "dataSources.properties.schema.json#/definitions/button" }] },
+                    "right": { "allOf": [{ "$ref": "dataSources.properties.schema.json#/definitions/button" }] },
+                    "down": { "allOf": [{ "$ref": "dataSources.properties.schema.json#/definitions/button" }] },
+                    "up": { "allOf": [{ "$ref": "dataSources.properties.schema.json#/definitions/button" }] }
```
toji: In my experience with traditional gamepads it's not uncommon for Dpads to be exposed as "digital axes", in that they're axis data that only reports -1, 0, or 1. If a UA was reporting device values directly rather than attempting some form of internal mapping you may end up needing to expose a pair of axes as a semantic Dpad.

schemas/responses.properties.schema.json
```diff
@@ -0,0 +1,36 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "id": "https://immersive-web/gamepad-mapping/0.1.0/responses.properties.schema.json",
+    "type" : "object",
+    "definitions": {
+        "target" : {
```
toji: Hm... thought from chat conversation: If you allowed multiple asset formats I don't know if it's possible to guarantee compatible node paths between all of them. You'd probably need another level of indirection where values like this indexed into an array of asset-specific paths. (Bleh.)

schemas/hand.schema.json
```diff
@@ -0,0 +1,35 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "id": "https://immersive-web/gamepad-mapping/0.1.0/hand.schema.json",
+    "type": "object",
+    "properties": {
+        "asset" : {
```
toji: Note related to chat: If you stick with a single asset identifier here then I advocate that it *must* be a standardized format.

schemas/mapping.schema.json
```diff
@@ -0,0 +1,37 @@
...
+    "type": "object",
+    "description": "The root object for a gamepad mapping.",
+    "additionalProperties": false,
+    "required": [ "version", "gamepad" ],
+    "properties": {
+        "version" : {
+            "description": "The schema version",
+            "type" : "string"
+        },
+        "gamepad" : {
```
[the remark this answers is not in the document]
NellWaliczek: That's actually what I had in the original design.  I changed it to this approach because I was simplifying the look-up process.  Developers know what folder to look for based on the gamepad id, which also makes it backwards compatible for WebVR :D

schemas/hand.schema.json
```diff
@@ -0,0 +1,35 @@
...
+            "uniqueItems": true,
+            "items": {
+                "allOf": [ { "$ref": "mapping.schema.json#/definitions/mappingIndex"} ]
+            }
+        },
+        "primaryButton": {
+            "allOf": [ { "$ref": "mapping.schema.json#/definitions/mappingIndex"} ],
+            "description": "The primary button suggested by the hardware manufacturer"
+        },
+        "primaryAxes": {
```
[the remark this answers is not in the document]
NellWaliczek: yep.  it's the index in the components array.  Duly noted that the name sux... got a suggestion?  primaryAxesComponent?

schemas/components.schema.json
```diff
@@ -0,0 +1,39 @@
...
+            },
+            "labelTransform" : {
+                "type": "string",
+                "description": "The name of the component's node in the asset file at which location a legend description can safely be placed"
+            },
+            "pressResponse" : {
+                "allOf": [ { "$ref": "mapping.schema.json#/definitions/mappingIndex"} ],
+                "description": "The index of the response to be invoked when the data source is pressed"
+            },
+            "touchResponse" : {
```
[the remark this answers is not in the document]
NellWaliczek: Not entirely sure I'm following the comment...Are you suggesting that a single response element should have the nodes for both touch and press?

schemas/dataSources.properties.schema.json
```diff
@@ -0,0 +1,47 @@
...
+                },
+                "supportsTouch" : {
+                    "description" : "Indicates all of the buttons is capable of reporting GamepadButton.touched",
+                    "type" : "boolean",
+                    "default" : "false"
+                },
+                "supportsPress" : {
+                    "description" : "Indicates all of the buttons is capable of reporting GamepadButton.pressed",
+                    "type" : "boolean",
+                    "default" : "true"
```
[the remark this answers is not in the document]
NellWaliczek: I'm a bit confused... supportsPress does default to true... were you asking me to remove it altogether?  How would you suggest representing the thumbrest then?

schemas/dataSources.properties.schema.json
NellWaliczek: There isn't a way to differentiate that?  Wouldn't the buttons array element just be null?

schemas/dataSources.properties.schema.json
NellWaliczek: Yep!  The original idea is that we could put optional min/max in there, but then I added the concept of the browser overrides and I realized it might be easiest to just mimic the format in the section below... I'm open to other approaches though...

schemas/dataSources.schema.json
NellWaliczek: For real?  Ugh.  I was patterning off of the gamepad spec which says they should be individual buttons....

schemas/responses.properties.schema.json
NellWaliczek: Bleh indeed.

schemas/hand.schema.json
NellWaliczek: And that's exactly why i was asking you that question in chat :)

schemas/components.schema.json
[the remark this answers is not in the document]
toji: I'm saying that the touch response element already does, so why does the differentiation at this level matter?

schemas/dataSources.properties.schema.json
toji: Hrm... I'm not sure that the element can be null. In order to allow that the Gamepad spec would have to indicate that it's a ` FrozenArray<GamepadButton?>`, but it currently specifies ` FrozenArray<GamepadButton>`. As far as I know there's no way that the gamepad API provides to indicate actual gaps in the button/axes mapping.

schemas/dataSources.schema.json
toji: Agreed that the browser should really be exposing them as button elements when it can, but in traditional Gamepad land we would run into a lot of cases where a device was exposed as a generic gamepad (not through, say, XInput or a similar well-structured API) and the best way to surface them was just to directly report the axes and buttons as they came off the device HID reports. This sometimes results in DPads that were reported as axes.

In general I expect this to be less of a problem with WebXR, because we have fewer possible sources of controller data and they are generally strictly typed, but it's not impossible that some UAs (or, more likely, custom branches of a UA) would do something that results in less rigorous UA-side mapping.

schemas/components.schema.json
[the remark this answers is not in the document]
NellWaliczek: oh!! Because it's not the same on a touchpad!  The touch response on a touchpad moves the little dot around.  the press response causes the whole touchpad to wobble.

schemas/dataSources.properties.schema.json
NellWaliczek: Erk... well @thetuvix isn't going to be happy to hear that....

schemas/mapping.schema.json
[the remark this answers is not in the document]
NellWaliczek: I've filed #5 to address this

schemas/hands.schema.json
```diff
@@ -0,0 +1,32 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "id": "https://immersive-web/gamepad-mapping/0.1.0/hands.schema.json",
+    "type": "object",
+    "description": "The hands",
+    "oneOf":[
```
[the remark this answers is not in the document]
NellWaliczek: Do you think we should support left+neutral and right+neutral as valid configurations?

schemas/dataSources.properties.schema.json
NellWaliczek: And yes, I agree that it should be `gamepadIndex` instead of `gamepadButtonsIndex` and `gamepadAxesIndex`

schemas/responses.properties.schema.json
NellWaliczek: Filed #6 to address this

NellWaliczek:
Note to self: update the license in this readme!

schemas/hands.schema.json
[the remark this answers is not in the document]
NellWaliczek: Per our phone call we'll allow neutral+left+right but not the ones in the question above

schemas/hand.schema.json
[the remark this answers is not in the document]
NellWaliczek: will use primaryAxesCompnoent and primayButtonComponent per our call

schemas/components.schema.json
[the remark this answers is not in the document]
NellWaliczek: talked this through on the call.  will add a response type to each element in the responses array and have the component have a single responses id array.  rename responses to visualResponses

schemas/dataSources.properties.schema.json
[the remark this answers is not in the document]
NellWaliczek: agreed to file an issue if i can't come up with a better inverted name before merging.

schemas/dataSources.properties.schema.json
NellWaliczek: decided the dpad edge case regarding button properties being inconsistent can be handled in the overrides file

schemas/dataSources.schema.json
NellWaliczek: modify the schema to allow for either xAxis/yAxis or 4 buttons in a dpad

NellWaliczek:
@toji  I'm not super happy with the current state of the `buttonsProperties` portion of the schema, so I'm going to poke at it a bit more.  But in the interest of having you look at the rest of this, I'm sending it your way now :)

NellWaliczek:
I'm strongly considering removing the "gamepad" property and just moving all of its children up a level.  WDYT?

NellWaliczek:
I'm also questioning the value of `touchSupportedAtMinValue` (formerly `touchSupported`)... I'm inclined to strike it for now and file an issue to consider adding it back in.

NellWaliczek:
Just pushed my last batch of significant changes.  With this, the README.md should now be updated to reflect the schema changes we discussed last week.  And given that there's a lot in here already, I'm inclined to only fix minor bug fixes in the current PR.  I'll file issues for any significant feedback and hop on it right after I get the tests PR merged.

README.md
```diff
@@ -1,454 +1,416 @@
...
-1. Find out the state of the controller's buttons and axes
-1. Modify the virtual model to reflect the state of the buttons and axes each render frame
-1. Provide a key or legend that explains the mapping of button/axes to actions
+1. Find out the state of the controller's component parts (made up of buttons and axes)
+1. Modify the virtual model to reflect the state of the components each render frame
+1. Provide a key or legend that explains the mapping of components to actions
+
+The state of an XR motion controller's buttons, thumbsticks, dpads, and touchpads is made available to developers via the the [Gamepad API](https://www.w3.org/TR/gamepad/). This data is divided up and populated in the `Gamepad.buttons` array and the `Gamepad.axes` array. While this system was adequate for the relatively homogenous console gaming controllers, it breaks down for XR motion controllers as they have not yet converged on a common form factor. In addition, the [Gamepad API](https://www.w3.org/TR/gamepad/) does not provide any information about the visualization of a `Gamepad` object which is a requirement to displaying a virtual copy of motion controller on opaque XR headsets.
```
toji: > The state... is made available to developers via the the Gamepad API.

Nit: That seems to suggest that it would be accessible via `navigator.getGamepads`. I suggest instead "A `Gamepad` object, as defined by the Gamepad API"

README.md
```diff
@@ -1,454 +1,416 @@
...
+                "yAxisIndex" : 1
             }
         ]
     }
 }
 ```
 
-### Responses
-Responses are the virtual manifestations of the physical motion controller's changes.  In order to generically reprsent the range of motion particular controllers allow for their various parts, the virtual models contain nodes representing these ranges. There are six currently defined types of virtual responses to physical motion.
+If the thumbstick or touchpad is able to be depressed, the data source must also include a `buttonIndex` property with a value representing the index in the `Gamepad.buttons` array at which to find the button's data. On touchpads with a dpad-like physical behavior, the x-axis and y-axis values can be used to determine which quadrant(s) are being pressed. For example:
```
toji: > If the thumbstick or touchpad is able to be depressed...

Maybe "If the thumbstick or touchpad is able to be depressed **or is touch sensitive**"? Even if it's not clickable we'd need the button (with `pressUnsupported: true`) to indicate touched state, right?

README.md
```diff
@@ -1,454 +1,416 @@
...
+
+Each `visualResponse` must contain a `target` property which references the node in an asset to be modified.  Each `visualResponse` must also contain a `userAction` property set to either `onTouch` or `onPress`.  A `userAction` set to `onTouch` indicates the visual response is describing the desired state of `target` when the associated `GamepadButton.touched` is `true` or the associated elements in the `Gamepad.axes` array are non-zero.  A `userAction` set to `onPress` indicates the visual response is describing the desired state of `target` when the associated `GamepadButton.pressed` is `true`. When `visualResponse` objects have the same `target`, node properties described by a `userAction` equal to `onPress` will take precedence over those with a `userAction` of `onTouch`.
+
+In addition to the `target` and `userResponse` properties, a `visualResponse` must have one of the following groups of properties.  Each property represents a node in the asset at the extent of the value which is it matched to:
```
toji: `userAction` instead of `userResponse` here?

toji:
> I'm strongly considering removing the "gamepad" property and just moving all of its children up a level. WDYT?

This doesn't seem like a terrible idea? Any idea what else might need to live at that level that we'd want to differentiate from the gamepad data? Otherwise, yeah, I'd support this. (Probably as a follow up PR.)

README.md
NellWaliczek: In the spirit of "no duplicate data", I'd been assuming any non-zero values on the xaxis or yaxis were indicators of a touched state and as a result no button index would be present if it was non-touchable.  Am i missing something?

NellWaliczek:
Filed #18 to track removing the "gamepad" property

README.md
toji: Not necessarily true, since on an Oculus Touch I can rest my thumb on the joystick without moving it, and the deadzone should be generous enough that it would still report `[0, 0]`.

To be clear, I think the schema and logic are right, I'm simply suggesting that this one sentence might need to be expanded. Either way it's a minor difference that doesn't affect any devices I'm aware of (Any that you can think of that have touch-capacitive axis controls that's aren't clickable?)

NellWaliczek: Ah, gotcha.  That's a good point about the deadzone.  And no, I don't know of any like that... I'll fix the sentence and merge.  thanks!!

NellWaliczek: BTW, this isn't just a prose fix.. the schema doesn't have `pressUnsupported` on that type of datasource.  Not a big deal to add it though

PR #3, opened 2019-02-07
Changing repo licence to MIT

NellWaliczek:


toji:
LGTM! WG/CG chairs were also consulted on this change with no objections.

PR #4, opened 2019-02-07
Making this repo a node package and adding tests with Jest

NellWaliczek:
This PR is the first step in showing how the schemas and mappings should be validated.  It also includes a MockGamepad class and a sample test file showing how it could be used.

Obviously there's still potential schema changes, but this PR sets the foundation for how we can iterate without the risk of breaking everything.  I'm also new to node and jest, so I'm guessing there's some further optimizations I can do so we don't have test files that are filled with copy paste stuff.  That's for a future PR though :)


NellWaliczek:
Closing in favor of #15 for now.  Functionality not in #15 will be posted in a subsequent PR

Issue #9, opened 2019-02-07
Investigate how to optimize schema tests


NellWaliczek:
As of PR #4, the schema tests are a lot of copy/paste.  I'd like to figure out how to get jest to consume a manifest of some sort and automatically run the same test against the manifest's supplied input.

NellWaliczek:
Resolved in #15

Issue #7, opened 2019-02-07
Add tests for `response` schema 


NellWaliczek:
Other schema portions have tests in the tests/schemas folder as of #4; add the response tests there.

NellWaliczek:
Closed in #15

PR #15, opened 2019-02-09
Node package and schema tests

NellWaliczek:
This PR replaces #4 and scopes the content down for ease of CR.  It now only does two things: turns the repo into a node package and adds tests to validate the schema.  It does not contain tests to validate the mapping files yet.  That will come in a subsequent PR.
Ignore the changes to README.md; my next step is to fix up that file as part of PR #2 and I'll rebase on top once that's done.


NellWaliczek:
Fixes #7

NellWaliczek:
Fixes #9

package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "gamepad-mappings",
```
toji: Color me surprised that this doesn't already exist, but even still I think something along the lines of "xr-gamepad-mappings" might be more appropriate?

package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "1.0.0",
```
toji: Is this at a functional-enough state to be considered version `1.0.0`? Don't really care what the version is, honestly, but people tend to think that "1.0" is a statement of stability/release-worthiness.

NellWaliczek: Ah typo. I'll fix it.  I meant to say 0.1.0

package.json
NellWaliczek: makes sense.... does that mean the repo name should be xr-gamepad-mappings?

toji: I don't know what npm etiquette is regarding that kind of thing, but probably?

PR #22, opened 2019-02-14
Changes visualResponse to contain an option for both press and touch

NellWaliczek:
This is the rough idea of the variant I proposed on the call today @toji .  I'm wanting to tweak some naming before I'd merge it, but what's your take on the general approach?

### DIFF
--- a/mappings/045E-065D/mapping.json
+++ b/mappings/045E-065D/mapping.json
@@ -83,48 +83,62 @@
     ],
     "visualResponses" : [
         {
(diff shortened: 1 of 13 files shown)


NellWaliczek:
Yep, you are correct in the conclusion that only touch or press can happen at the same time.  However, there may be different target nodes that need to be affected for a single component's state to be reflected as touched or pressed.  Hence the arrays.

In any event, This branch is now woefully out of date, so I'm going to close this PR in favor of the real one.

PR #23, opened 2019-02-20
Adds source code for interpreting mapping files

NellWaliczek:
This PR adds several things:
1. The initial version of some JavaScript source code that loads the appropriate mapping file for a given Gamepad object and enables XR devs to reason about the components rather than the Gamepad object.
2. The initial version of a mock gamepad implementation that initializes based on the data in a mapping file
3. A bunch of test helper functions for generating data objects for use in tests
4. Tests for # 1 that uses #s 2 and 3.  So. Many. Tests.

There's still more work to do here, but once this portion is merged my plan is to switch gears and focus on feathering in support for WebVR so we can start using this library with browsers that are already available (this includes getting a few sample asset files hooked up).

(yes, I know I'm on vacation... I just really want this done!)


src/constants.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  Handedness: Object.freeze({
+    NEUTRAL: "neutral",
```
toji: FYI: We landed the API-side change to "none" on Tuesday when there were no objections on the call

src/xrGamepad.js
```diff
@@ -0,0 +1,76 @@
...
+  constructor(gamepad, mapping, handedness) {
+    if (!gamepad || !mapping) {
+      throw new Error("Cannot create XRGamepad from invalid parameters");
+    }
+
+    if (gamepad.id != mapping.id) {
+      throw new Error(`Gamepad id ${gamepad.id} and mapping id ${mapping.id} do not match`);
+    }
+
+    if (gamepad.handedness && gamepad.handedness != handedness) {
```
toji: What's the benefit to passing in the handedness here, as opposed to just picking up the one that's reported by the gamepad? This largely seems like it's just forcing developers to pass in gamepad.handedness again and slapping them on the wrist if they don't?

src/constants.js
NellWaliczek: Yeah I figured I'd handle that in a separate PR if that's ok

src/xrGamepad.js
NellWaliczek: WebXR won't be repording handedness as part of the gamepad, yes?  It's just WebVR that does that.  I could change the check to make handedness optional and only error out if it's not on the gamepad...wdyt?

toji: Ah, yes. I'd forgotten about the difference in WebVR/XR here. Don't mind me! This is fine.

NellWaliczek:
Spoke a bit offline and was talked out of splitting the CR into 4 separate ones.  ^_^ I agree this should be the last of the large changes to seed the repo.  From this point on the PRs should be far more manageable in size.

PR #31, opened 2019-02-25
Updates the readme to match the latest source/schema changes

NellWaliczek:
* Removes the "gamepad" node
* Updates the visualResponses section to match the new format
* Adds a description of the top-level properties "version" and "id"

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,7 +346,7 @@ For example:
     "handedness" : {
         "none" : {
             "asset" : "some-url",
-            "root" : "neutral-handedness-controller",
+            "root" : "none-handedness-controller",
             "components" : [0]
         }
     }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -353,14 +343,11 @@ For example:
...
-        "handedness" : {
-            "none" : {
-                "asset" : "some-url",
-                "root" : "none-handedness-controller",
-                "components" : [0]
-            }
+    "handedness" : {
+        "none" : {
+            "asset" : "some-url",
+            "root" : "neutral-handedness-controller",
```
toji: This looks like it was changed to "none-handedness-controller" below. Applicable here as well, I assume?

README.md
```diff
@@ -370,21 +357,28 @@ The `none`, `left`, and `right` objects may also contain two additional properti
...
+            "components" : [0],
+            "primaryButtonComponent" : 0,
+            "primaryAxisComponent" : 1
         }
     }
 }
 ```
 
+## Top-level Properties
+The schema defines two key top-level properties: `version` and `id`.  The `version` property is a semantic version made up of a major, minor, and patch version number.  The major and minor version parts must always be up to date with the library being used to parse the mapping file.  The patch portion of the version may be out of sync.  The `id` property must match the `Gamepad.id` of the `Gamepad` object the mapping is being applied to.  The naming conventions for the `WebXR Device API` are concretely defined; the naming conventions for `WebVR` are somewhat more haphazard.
```
toji: > The schema defines two key top-level properties: `version` and `id`.

There's more top level properties than this, though, right? Especially since everything in the readme has been moved up a level in this PR.

NellWaliczek: I also corrected the version language to match what we decided on yesterday

Issue #5, opened 2019-02-07
Ensure schema can be used for both WebVR and WebXR


NellWaliczek:
It could be that the way to do this is to have two separate folder hierarchies?  This would allow us to have the "Oculus Touch (left)" for WebVR and the correct "OculusTouch" for WebXR. Alternatively, we could put another .json file at the root of the repo containing a list of reported gamepad IDs and which folder contains the mapping/asset.

Also, we'll need to add a component node for the target ray origin for WebVR.  

Anything else?

NellWaliczek:
Experimenting with options in https://github.com/immersive-web/Gamepad-mappings/tree/webvr-and-overrides

PR #34, opened 2019-02-26
Adds support for WebVR mappings

NellWaliczek:
This change moves existing mappings into a WebXR subfolder and renames
them to conform with WebXR naming conventions. It then adds a WebVR
subfolder in which new WebVR mappings are added based on the known
gamepad ids in released browsers. The schema is extended to add an
property indicating a mapping is for WebVR. When this property is true,
additional WebVR-specific properties for targetRay origin and axis
inversion are added to the valid schema.
Tests are added and test tools are modified to look for  mapping files
in both WebVR and WebXR subfolders.


NellWaliczek:
Fixes #5

tests/testHelpers.js
```diff
@@ -29,9 +31,9 @@ const TestHelpers = {
    * be searched instead of the real mappings folder
    * @returns {Object} The mapping described in the mapping.json file
    */
-  getMappingById : function (gamepadId, useMocks) {
-    let folder = useMocks ? mockMappingsFolder : mappingsFolder;
-    let mappingPath = join(folder, gamepadId, "mapping.json")
+  getMappingById : function (gamepadId, mappingType = Constants.MappingType.WEBXR) {
```
NellWaliczek: Fix function metadata

mappings/WebVR/Oculus Go Controller/mapping.json
```diff
@@ -0,0 +1,99 @@
+{
+    "version" : "0.1.0",
```
toji: I assume you want these to be just "0.1" to match with #35?

toji:
Didn't verify all the mappings 😉 but the rest looks good with one question about the version.

mappings/WebVR/Oculus Go Controller/mapping.json
NellWaliczek: Yeah, I wasn't sure which would get approved first ;)

PR #38, opened 2019-02-26
Adds explanation to readme of how to use the javascript library

NellWaliczek:


README.md
```diff
@@ -431,6 +431,97 @@ Certain WebVR `Gamepad` objects have some components with an inverted `yAxis` ca
...
+      const weightedVisualizations = xrComponent.getWeightedVisualizations();
+      MyEngine.updateInputVisuals(key, weightedVisualizations);
+    });
+  });
+}
+```
+
+> TODO Add an explanation for how to updated the visuals based on the weighted visualizations
+
+## Target Ray Origin
```
toji: So this is feedback that's a couple of PRs too late, but it didn't really click with me till I saw it used in context. The use of "origin" here feels to me like it only implies a position offset, so it read kind of weird to me at first. Is there a better term that we could use?

README.md
```diff
@@ -431,6 +431,97 @@ Certain WebVR `Gamepad` objects have some components with an inverted `yAxis` ca
...
+In WebXR, the `XRInputSource.targetRaySpace` should be passed into `XRFrame.getPose()` to determined the pointing origin.
+
+```js
+const targetRayOrigin = xrFrame.getPose(xrInputSource.targetRaySpace, xrReferenceSpace);
+```
+
+### WebVR Target Ray Origin
+In WebVR there is no implicit mechanism for retrieving a target ray origin.  Instead, it must be retrieved from the the mapping via the `XRGamepad`
+```js
+const targetRayOrigin = xrGamepad.targetRayOrigin;
```
toji: Hm... maybe I just misread this previously: I thought that `targetRayOrigin` was a static offset from the base gamepad pose? That would appear to not be the case here if this is meant to be ~equivalent to the WebXR code above.

README.md
NellWaliczek: I'm open to whatever, though I'd like to not hold this PR for it...

README.md
NellWaliczek: Gah.  No that's me moving too fast.  I'll get an update pushed in a moment.

PR #61, opened 2019-08-02
Improvements to the manual test

NellWaliczek:
This PR adds a bunch of stuff:
* Dropdowns to select from available profiles
* In watch mode, the build detects profile changes
* Query string parameters to select the initial profile to load
* Better error messaging when an asset is not found
* Fixes a bug in how MockGamepads are generated

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -918,12 +918,6 @@
       "integrity": "sha1-7GphrlZIDAw8skHJVhjiCJL5Zyo=",
       "dev": true
     },
-    "async-each": {
-      "version": "1.0.1",
-      "resolved": false,
-      "integrity": "sha1-GdOGodntxufByF04iu28xW0zYC0=",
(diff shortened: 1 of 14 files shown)


NellWaliczek:
No objections to doing so.  There's two things that will need to happen first, though:
* The test page needs to link off to the readme so people can find their instructions easily when that's the repo's hosted page
* As is, the page isn't super useful TBH.  It doesn't look for real hardware if it's available, nor does it allow people to test hardware that isn't already in the list.  I'm thinking of either adding the ability to point to an arbitrary hosted location to load profiles or to add support for drag/drop.  In both cases, I'd like to be thoughtful about the iteration loop for testing new assets because it involves a lot of refreshing the page (hence my adding query string parameters in this PR).

So, basically, I didn't want to jam anymore stuff into this PR ^_^

Issue #59, opened 2019-07-25
Host the test page


NellWaliczek:
The manual test should be updated and hosted from this repo so that new hardware can be tested easily

NellWaliczek:
Before this happens:
* The test page needs to link off to the readme so people can find their instructions easily when that's the repo's hosted page
* As is, the page isn't super useful in hosted form because it doesn't look for real hardware if it's available, nor does it allow people to test hardware that isn't already in the list. To fix this, we need do one or both of the following.  In doing so, we need to be thoughtful about the iteration loop for testing new assets because it involves a lot of refreshing the page.
  * Add the ability to point to an arbitrary hosted location to load profiles
  * Add support for drag/drop of asset files into the page

PR #63, opened 2019-08-06
Moved the profile viewer to an accessible place

toji:
Also renamed from "manualTest" to "viewer", linked from the README,
fixed a couple of stray links and the README title to match the repo,
and moved the required subset of Three.js files into the viewer
directory so that we don't have to check in node_modules.

Partially addresses #59, but only the hosting aspect.


README.md
```diff
@@ -23,10 +23,15 @@ This repository has been designed to meet the following goals:
...
 ## Adding New Hardware
-Fill in the steps for adding a folder for a new XR device, testing the change, and submitting for PR https://github.com/immersive-web/xr-gamepad-mappings/issues/51
+Fill in the steps for adding a folder for a new XR device, testing the change, and submitting for PR https://github.com/immersive-web/webxr-input-profiles/issues/51
 
 ## Filing A Bug
-Fill in the steps for filing a bug https://github.com/immersive-web/xr-gamepad-mappings/issues/52
+Fill in the steps for filing a bug https://github.com/immersive-web/webxr-input-profiles/issues/52
+
+## Profile Viewer
+This repository also contains a page that allows for easy testing and viewing of the devices described by each profile. It renders the assets described by the profile and allows emulated manipulation of each of the inputs in order to see the rendered response.
```
NellWaliczek: Can we add a link to issue #59 and call out the current test page's limitations?

NellWaliczek:
Oh, and perhaps we can make the name "profileViewer" because otherwise it makes me think of the viewer reference spaces ^_^'

profileViewer/modelViewer.js
```diff
@@ -21,6 +21,13 @@ function addTouchDots({ motionController, rootNode }) {
     if (component.dataSource.dataSourceType === 'touchpadSource') {
       // Find the node to attach the touch dot.
       const componentRoot = rootNode.getObjectByName(component.rootNodeName, true);
+
+      if (!componentRoot) {
+        // eslint-disable-next-line no-console
+        console.error(`Could not find root node of touchpad component ${component.rootNodeName}`);
```
NellWaliczek: Can we set this text in the assetErrorElement too?  Would make it easier for artists.  Same in the other places you're logging errors.  I'd be happy for this to be a helper function too.

profileViewer/modelViewer.js
```diff
@@ -137,6 +151,9 @@ function animationFrameCallback() {
     Object.values(activeModel.motionController.components).forEach((component) => {
       const componentNodes = activeModel.nodes[component.id];
 
+      if (!componentNodes)
```
NellWaliczek: Add a comment indicating that this can be skipped because errors are notified at load time?

rollup.config.js
```diff
@@ -17,7 +18,8 @@ export default [
       eslint(),
       copy(
         [
-          { files: 'profiles/**', dest: `${DIST_FOLDER}/profiles` }
+          { files: 'profiles/**', dest: `${DIST_FOLDER}/profiles` },
+          { files: 'node_modules/three/**', dest: `${VIEWER_FOLDER}/three` }
```
NellWaliczek: I guess we could copy just the required files if you'd like.

toji: Since it's just going to gh-pages I guess I don't care that much. Just makes the build noisier, but it would be more error prone to have to maintain the list of files that are in use.

toji:
Okay, updated again with some better error reporting handling all around.

rollup.config.js
NellWaliczek: really noisy.  Perhaps instead we can add a second copy of the plugin that isn't verbose for these files?

profileViewer/modelViewer.js
```diff
@@ -194,6 +206,8 @@ const ModelViewer = {
   initialize: () => {
     canvasParentElement = document.getElementById('modelViewer');
     assetErrorElement = document.getElementById('assetError');
+    assetErrorList = document.createElement('ul');
```
NellWaliczek: Why not not just make the assetError element be a UL instead of a div so you don't need to manage it here?

toji: Ah, yes, well you see, that was for the very good reason that I, in fact, did.... not.... think of that. 😅

PR #64, opened 2019-08-07
Extends Profile Viewer to allow pointing at a local folder

NellWaliczek:
This PR includes a handful of quality of life improvements.  
* profileViewer page is now able to load files from a local folder to test from. In Firefox, the selected file list is persisted on refresh; in Chrome it is not.  In either case, the 'custom' profile value is saved on the query string.
* Moves the profileViewer to deploy into its own distribution folder so that the build doesn't copy three.js into a source folder.
* Three.js file copies during the build are no longer verbose
* Added a file to manage error logging and display across all JavaScript files in the profileViewer
* Cleaned up the CSS so that errors are always legible and the error box is properly hidden when no errors are present
* Makes sure travis runs linting across all files not just tests and those in src/
* Adds a package.json script to simulate running travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ cache:
 script:
   - npm run build
   - npm run test
+  - npm run lint
 
 deploy:
   provider: pages
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,7 @@ cache:
     - node_modules
 
 script:
-  - npm run build
-  - npm run test
-  - npm run lint
+  - npm run travis
 
 deploy:
   provider: pages
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 dist/
+profileViewerDist/
--- a/package.json
+++ b/package.json
@@ -43,12 +43,12 @@
     "MotionControllers"
   ],
   "scripts": {
-    "clean": "rm -r dist && rm -r profileViewerDist",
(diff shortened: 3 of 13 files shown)


NellWaliczek:
Fixes #59

.travis.yml
```diff
@@ -8,8 +8,7 @@ cache:
     - node_modules
 
 script:
-  - npm run build
-  - npm run test
+  - npm run travis
```
toji: What's the benefit of collapsing all the steps into one like this? Travis is failing and I'm not really in a position to dig into why, but I think most of the tooling would give a more descriptive error about which step failed if they were all listed separately here.

NellWaliczek: So that I can be sure in advance of pushing code that I've done my due dilligence and travis won't fail.  Now, in this case I made a mistake in my "clean" script and only didn't notice it because the folder I was cleaning still existed.  In the travis run, the folder wasn't there because it was already clean.  I've got a fix for that locally, and I'll push that as soon as I've ensured all the files that need to be linted will be linted (and linted only once).

As for tooling giving descriptive errors, from what i can tell the errors are all laid out because it's just an "&&" of a bunch of separate commands that are run in order.

NellWaliczek: I'm gonna leave this for now, but I'm happy to discuss more when you're back.  In the meanwhile, I'm going to merge to unblock the artists.

PR #65, opened 2019-08-08
Remove the special case for button min/max node name defaults

NellWaliczek:
Nothing much more to explain here.  VisualResponse min/max nodes always default to "MIN"/"MAX" and the non-standard node names have been added to the profile.json for the windows controllers.


NellWaliczek:
Merged to unblock technical artists.  If there are issues with this, I'm happy to fix them later.

PR #66, opened 2019-08-09
fix: got linting work with rollup

KaboomFox:
```globals - An array of global variables to declare (default: empty array). Corresponds to --global, and similarly supports passing 'name:true' to denote a writeable global. Note: This differs from .eslintrc.* / baseConfig, where globals is an object.``` -https://github.com/eslint/eslint/blob/master/docs/developer-guide/nodejs-api.md

I created an issue here: https://github.com/TrySound/rollup-plugin-eslint/issues/31

I took the globals out for the time being.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,3 @@
 dist/
+node_modules/
+profileViewer/
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,10 +6,6 @@
     "extends": [
         "airbnb-base"
     ],
-    "globals": {
-        "Atomics": "readonly",
-        "SharedArrayBuffer": "readonly"
-    },
     "parserOptions": {
         "ecmaVersion": 2018,
         "sourceType": "module"
--- a/package-lock.json
+++ b/package-lock.json
@@ -750,6 +750,12 @@
         "fastq": "^1.6.0"
       }
     },
+    "@types/eslint-visitor-keys": {
+      "version": "1.0.0",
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: .eslintignore [hunk 1]
```diff
@@ -1 +1,3 @@
-dist/
\ No newline at end of file
+dist/
+node_modules/
```
NellWaliczek: eslint is supposed to ignore this folder by default

.eslintignore
```diff
@@ -1 +1,3 @@
...
+profileViewer/
```
NellWaliczek: This is actually the folder I was bemoaning *wasn't* getting linted ^_^;

.eslintrc.json
```diff
@@ -6,10 +6,6 @@
     "extends": [
         "airbnb-base"
     ],
-    "globals": {
```
NellWaliczek: So this weirdness is totally a bug, but as you noted, it's just a bug with passing the config via the cli.  If eslint loads the config itself from the known config location, the bug isn't hit.  Coincidentally, the rollup-pluglin-eslint extra options can't be passed if the config is specified (it can only be either a config name string or an object), even though they have nothing to do with one another.  So instead, I ended up falling back to the default config location (which actually works fine) and passing in the 'throwOnError:true' config parameter to ensure the rollup build will actually fail when linting fails.

NellWaliczek:
Issue addressed by #64 instead

PR #67, opened 2019-08-09
Move all source files into src/ and change the profile viewer to distribute in the profileViewer folder

NellWaliczek:
The repo file structure was a bit unclean with a bunch of stuff sitting in the root directory.  This change moves all source files (profiles, schemas, tests, profileViewer etc) inside an new src directory.  It also changes the build target of the profile viewer to be the profileViewer folder.  (Note: the profileViewer folder previously contained the source for the viewer, but this has been moved to src/profileViewer)


NellWaliczek:
Again, I don't mean to be all YOLO by merging these changes without a CR.  Given that they're just folder changes and the build has passed, I'm going to take the risk.  If there are issues with this choice, @toji,  I'm happy to revisit once you're back from your time off.

toji:
Yeah, this feels much cleaner to me. Post-merge LGTM!

PR #74, opened 2019-08-14
Fix issues for Oculus Touch, HTC Vive, and WMR controller entries

jacobcdewitt:
Update the profiles for these controllers to reflect the order that buttons and input axes should be exposed to the JS with the latest xr-standard Gamepad mapping, which is explained [here](https://github.com/immersive-web/webxr-gamepads-module/blob/master/gamepads-module-explainer.md).

Also remove the entry for the menu button from the microsoft-045e-065d profile (which happens to be the Samsung Odyssey) because the spec says buttons reserved by the UA and/or platform must not be exposed.

### DIFF
--- a/packages/profiles-registry/profiles/htc-vive/profile.json
+++ b/packages/profiles-registry/profiles/htc-vive/profile.json
@@ -44,13 +44,6 @@
     ],
     "dataSources" : [
         {
-            "id" : "trackpad",
-            "dataSourceType": "touchpadSource",
-            "xAxisIndex" : 0,
-            "yAxisIndex" : 1,
-            "buttonIndex" : 2
-        },
-        {
             "id" : "trigger",
             "dataSourceType": "buttonSource",
             "buttonIndex" : 0,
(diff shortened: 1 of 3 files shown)


jacobcdewitt:
I removed the entry from the menu button for Windows controllers because I don't think it should be exposed to the JS. For example, Chrome does not expose the Windows menu button on WebXR gamepads.

This is based on the spec text "Buttons reserved by the UA or platform MUST NOT be exposed." from https://raw.githubusercontent.com/immersive-web/webxr-gamepads-module/master/index.bs

jacobcdewitt:
@toji - Can you double-check that we shouldn't be exposing a menu button for controllers?

NellWaliczek:
@thetuvix, can you weigh in on the menu button?  In the WebVR timeframe, it was exposed to developers and I'd like to confirm this change is intentional

NellWaliczek:
Hey @jacobcdewitt can you rebase this on top of #78 if anything still needs fixing?

jacobcdewitt:
> Hey @jacobcdewitt can you rebase this on top of #78 if anything still needs fixing?

I looked at #78 and I see it fixes the button ordering issues for the controllers updated in this pull request. However, per my comments in that pull request, I'm not sure if we should have any entries for menu buttons. That can be addressed in #78 , so I'm closing this one.

PR #75, opened 2019-08-16
Export the MotionController class

NellWaliczek:
In preparation for some other upcoming changes, I removed the static Profiles.createMotionController() function that fetches a profile and generates a MotionController.  Among other issues with it, this function was relying on the fact that Profiles.fetchProfile() was modifying the returned profile to include the assetUrl.

Instead, the MotionController class is exported from the library and the Profiles file now has a convenience function that computes the assetUrl for a profile/handedness combination (under the assumption that the profile was fetched using the same instance)

In addition this PR includes:
* Tests and test changes based on the above
* Refactoring the profileViewer to make it easier to introduce new mechanisms for retrieving profiles
* Fixed a mistake just introduced in the .eslintignore that was causing the whole profile viewer not to be linted


NellWaliczek:
Yeah, that's why I moved most of it into the handednessSelector.js file.  But I agree, it's still bugging me a bit too...

Issue #70, opened 2019-07-19
Several default profiles are ambiguous about the presence of a button


NellWaliczek:
While working on updating to the xr-gamepad-mappings library to match the latest spec text, I discovered a fairly annoying profile definition issue.  I'm expecting to have the library updates posted shortly, but in the meantime I figured I'd get this issue filed for discussion.

The `XRInputSource.getProfiles()` call is designed such that developers can iterate over the list returned and select the best profile they know how to render and interpret.  We've defined several default profiles to make this easier such as `button-controller`, `thumbstick-controller`, `touchpad-controller`, and `touchpad-thumbstick-controller`.  The trouble is we defined the latter three to be ambiguous about whether or not a button is also present.  In order to utilize this optional button or render its movement appropriately, developers must also check if the gamepad's mapping is `xr-standard` and potentially load/animate a different renderable model.

I know it's not ideal to have to define a bunch of additional default profile names (`button-thumbstick-controller`, `button-touchpad-controller`, `button-touchpad-thumbstick-controller`), but I'm not sure what other better approach there is.  I'm super open to ideas though!

/cc @thetuvix

thetuvix:
Has anyone done the exercise to produce the mappings from today's real-world controllers to the default profiles they can service?  Perhaps in practice past a certain complexity controllers always end up having at least one button.  I'm actually having trouble thinking of any controlles that have either a thumbstick or touchpad and no app buttons, and so we could define those default profiles to assume a button as well.

If such a controller really did come along down the line with those controls but no button, it's unlikely that today's content would have coded defensively to support it anyway, which is the whole point of being explicit about profiles - we can add a new profile then for sites that now have a real controller to reason about.

NellWaliczek:
I've got an artist building the assets for this repository right now.  And we'd be fine with requiring a button be part of a controller that reports 'thumbstick-controller', 'touchpad-controller', or 'thumbstick-touchpad-controller'.  The key thing is that we don't want controllers reporting one of those strings if no button is present which is something not clearly stated in the definition of those profiles.
And as you say, in the future we could add new profiles such as 'thumbstick-no-button-controller' if necessary.

jacobcdewitt:
[...]

PR #76, opened 2019-08-16
Potential registry format

NellWaliczek:
A really early stab at how we might define the registry.  This will probably end up moving into a different folder, but for now...

I put three sections in the json.  The first, is the list of agreed up on generic profiles.  The third is the legacy controller names for WebVR.  

The interesting part is the second section.  "customProfiles" is a terrible name, but in the interest of not getting stuck in the weeds bikeshedding.... the important bit here is that I've got an entry for each hardware device.  That entry's value is an array of the strings.  The XRInputSource.profiles[0] will be the custom profile key and all the subsequent XRInputSource.profiles elements will be the values of the custom profiles array in order. 

A few questions that come to mind...
* Do we want to make a requirement that each customProfile entry must end in a generic profile?  
* Do we want to make a requirement that there only be one generic profile in each custom profile's list?
* Should we remove the webvr legacy elements?
* Should we not bother separating the generic profiles at all?

Lots to think about.

### DIFF
--- a/src/profiles/registry.json
+++ b/src/profiles/registry.json
@@ -1,0 +1,28 @@
+{
+    "genericProfiles" : [
+        "button-controller", 
+        "thumbstick-controller", 
+        "touchpad-controller", 
+        "thumbstick-touchpad-controller"
+    ],
+    "customProfiles" : {
+        "google-daydream" : [ "touchpad-controller" ],
+        "htc-vive" : [ "touchpad-controller" ],
+        "microsoft-045e-065d" : [ "thumbstick-touchpad-controller" ],
+        "microsoft-045e-065b" : [ "microsoft-045e-065d", "thumbstick-touchpad-controller" ],
+        "oculus-go" : [ "touchpad-controller"],
+        "oculus-quest" : [ "oculus-touch", "oculus-go", "thumbstick-controller" ],
+        "oculus-touch" : [ "thumbstick-controller" ],
+        "samsung-gearvr" : [ "touchpad-controller" ],
+        "valve-index" : [ "htc-vive", "touchpad-controller" ]
+    },
+    "legacyWebVRProfiles" : [ 
+        "Daydream Controller",
+        "Oculus Go Controller", 
+        "Oculus Touch (Left)", 
+        "Oculus Touch (Right)", 
+        "OpenVR Gamepad",
+        "Spatial Controller (Spatial Interaction Source) O45E-065B",
+        "Spatial Controller (Spatial Interaction Source) O45E-065D"
+    ]
+}


src/profiles/registry.json
```diff
@@ -0,0 +1,28 @@
+{
+    "genericProfiles" : [
+        "button-controller", 
+        "thumbstick-controller",
```
jacobcdewitt: As discussed in issue #70 , these names are ambiguous about the presence of a grip button. Here is what I have implemented in Chrome, which avoids the ambiguity:
controller
touchpad-controller
thumbstick-controller
touchpad-thumbstick-controller
grip-controller
grip-touchpad-controller
grip-thumbstick-controller
grip-touchpad-thumbstick-controller

This covers all the possible xr-standard Gamepads by specifying whether or not each optional element (grip, touchpad, thumbstick) are present. The trigger button will always be present, so it's not required in the name. JS code could even parse these generic strings to figure out the basic functionality of the controller without even needing to query a registry.

src/profiles/registry.json
```diff
@@ -0,0 +1,28 @@
...
+    "genericProfiles" : [
+        "button-controller", 
+        "thumbstick-controller", 
+        "touchpad-controller", 
+        "thumbstick-touchpad-controller"
+    ],
+    "customProfiles" : {
+        "google-daydream" : [ "touchpad-controller" ],
+        "htc-vive" : [ "touchpad-controller" ],
+        "microsoft-045e-065d" : [ "thumbstick-touchpad-controller" ],
```
jacobcdewitt: This happens to be the "Samsung Odyssey", which could be exposed with a human readable name of "samsung-odyssey".

src/profiles/registry.json
toji: I'm supportive of making the presence of a grip button part of the generic profiles. Initially I was worried about a combinatorial explosion of profile strings, but if we limit it to just the `xr-standard` inputs and list them in the same order each time (as Jacob shows) then it seems very manageable and useful to developers. I have no opinon on the ordering of elements other than to say I'd be fine with the above list as-is. (With the addition of `button-controller` as we originally had.)

jacobcdewitt: Is `button controller` supposed to mean a controller with just a trigger button (but no grip button)?

And yes, I agree we should limit the generic profiles to covering functionality from xr-standard inputs. If developers want to know about additional optional buttons (such as Oculus touch A/B buttons), then they need to look at the more specific profile strings.

toji: Yes, `button-controller` would be a device with only a trigger button (or only a primary button, even if it wasn't trigger-shaped.) It would basically be the fallback for an `xr-standard` device where none of the other generic profiles applied. The only device I'm aware of that it would apply to off the top of my head is the [HoloLens clicker](https://docs.microsoft.com/en-us/windows/mixed-reality/hardware-accessories#hololens-clicker), but regardless of how common it is it's a decen idea to have a clear plan for how to address any devices that fit that description.

NellWaliczek:
In the realm of "huh, that's an eerie coincidence", I'd actually started to ponder this exact problem earlier today.  It's also somewhat related to this issue I filed in the gamepad repo a while back https://github.com/immersive-web/webxr-gamepads-module/issues/3.

In any event, after talking it over with @toji a bit, I'd like to propose the following generic profile names.
```
        "xr-unknown",
        "xr-standard", 
        "xr-standard-touchpad", 
        "xr-standard-thumbstick", 
        "xr-standard-touchpad-thumbstick",
        "xr-standard-grip",
        "xr-standard-grip-touchpad", 
        "xr-standard-grip-thumbstick", 
        "xr-standard-grip-touchpad-thumbstick"
```

Our reasoning behind the names was based on several factors.  
1. The hardware-based (non-generic) profile names all start with a vendor id and the generic ones did not.  Now it's not super likely that a company named "touchpad" would start shipping XR hardware, but I guess you never know.  
2. All xr-standard controllers are required to support a primary button so we didn't think it necessary to call it out specifically.  All other allowable combinations of additional components are covered by items in this list.
3. It is allowed to have an XRInputSource.gamepad even if it isn't xr-standard, so it's possible that we'll need at least one string to indicate that situation.  However, it is also possible that when we address https://github.com/immersive-web/webxr-gamepads-module/issues/3 that this will no longer be necessary, but i'm not entirely sure.

Lastly, given that this list of generic profiles ended up fairly tightly tied to the xr-standard mapping, I'd like to suggest we move the definition of these particular generic profiles into the webxr-gamepads-module.  To reiterate something we've said before though, the profile ids are not unique to gamepad-based input sources.  Which means we'll need to make it very clear in the module text that the gamepad-related generic profiles list does not exhaustively cover all potential XRInputSources, but that it only exhaustively covers profileIds allowed by an XRInputSource with a non-null gamepad.

src/profiles/registry.json
NellWaliczek: During the WebVR timeframe, these were the strings that Microsoft chose to expose.  @thetuvix needs to weigh in on what Microsoft actually wants here.

jacobcdewitt:
ping ...

NellWaliczek:
Closing in favor of #78

Issue #71, opened 2019-04-02
Define a registry for defining how UAs expose input devices


ddorwin:
While the WebXR spec (after immersive-web/webxr#553 is merged) provides some guidelines around exposing controllers as Gamepads, there is still room for interpretation that could lead to different behavior among implementations. In addition, future controller/input mechanisms (i.e., hands as exposed by a specific technology) may not map well to `"xr-standard"` or otherwise have obvious mappings.

To encourage consistency and interoperability, details about how specific controllers should be exposed by user agent implementations should be defined in a registry. The registry would be separate from the spec, allowing out-of-band updates (per an established process). It would both clearly define how implementations should expose specific controllers and thus how applications and libraries (e.g., https://github.com/immersive-web/xr-gamepad-mappings) can expect them to be exposed.

The registry would map some well understood definition of a physical controller to:
* `id` string
* `GamepadMappingType`
* The physical button that represents the [primary action](file:///src/scratch/webxr/index.html#primary-action)
* A clear definition of how the physical buttons are exposed via the `Gamepad` interface
  * This would include which buttons are not to be exposed, such as those reserved by the system and/or user agents.

Ideally, the entry would also include some concrete information, such as USB vendor and product IDs and/or identifier exposed by native runtime(s).

A single entry could cover multiple devices (i.e., revisions of a controller) as long as the physical appearance (due to the [intent](https://github.com/immersive-web/webxr/issues/336#issuecomment-478712300) for applications to use the `id` to choose a controller model) and button mapping are the same. If there are multiple versions of hardware with the same product name, they may be identified by version, year, or similar.

Such a registry would also solve a couple open issues:
* Format of Gamepad identifiers (#550): While the registry could have guidelines, there wouldn’t be a need to define a perfect rule covering all hardware because the registry would specify the identifier.
* As such, identically-/similarly-named controllers (e.g., “Oculus Touch” per https://github.com/immersive-web/webxr/issues/550#issuecomment-469818569) can be clearly and easily differentiated.

<br/>One additional possibility is that [XR-specific] GamepadMappingTypes also be specified in a registry. These would be added much more conservatively, but it might be preferable to be able to define new ones without going through the full spec process. This would also allow slimming the core WebXR spec a bit.

<br/>For reference, there is an open issue for the W3C Process Document (https://github.com/w3c/w3process/issues/168) to define a process for handling registries. This shouldn’t block us from proceeding, and other specs have successfully used registries in the past.

[...]

Issue #54, opened 2019-07-25
README.md - Fill in supported hardware list


NellWaliczek:
In the README.md, the Supported Hardware section in the appendices needs to be filled in.  Instructions for keeping this up to date should be included in #51

NellWaliczek:
This has been moved into the registry package.

Issue #26, opened 2019-02-23
Decide on versioning semantics


NellWaliczek:
In the context of a major/minor/patch version scheme, we need to decide which types of changes to this repo would trigger a version bump.  Throwing down a first proposal, I suggest:

**Patch**
* Adding new mapping and asset files
* Fixes to existing mapping or asset files

**Minor**
* Fixing bugs in source code
* Fixing bugs in tests
* Non-breaking schema changes (additive features)
* Non-breaking source changes (additive features)

**Major**
* Breaking schema changes
* Breaking source changes

Additionally, I'm wondering what we want to do about the version property in the mapping files.  It seems a bit much to rev the version # in a mapping file simply because a new mapping file was added.  Perhaps we change the version property in the mapping.json files to just be major/minor?

NellWaliczek:
Based on several discussions over the past few days, I'm updating the proposal to as follows.  The key aspect of this proposal is that mapping.json file versions will only update when a major/minor version bump occurs.

#### Patch
* Adds new mapping and/or asset files
* Fixes to existing mapping and/or asset files
* Critical security fixes
* Occasional non-breaking fixes to schema, source code, or test code

#### Minor
* Additional features added 
* Small breaking changes to schema
* Breaking changes in source code or test code

#### Major
* Significant design changes

NellWaliczek:
Related to this, what should the branch structure be to allow automatic publishing when a new release is generated?

toji:
Added Nell's breakdown of the versioning from this issue to the repo's README, since I believe it's still an accurate representation of how we intend to manage the package versioning.

PR #77, opened 2019-08-28
Refactoring into a monorepo that allows for publishing multiple packages

NellWaliczek:
This PR doesn't change any functional behavior of the files in the library.  It moves the logical folders into a packages directory that lerna can use to manage dependencies and simplify publishing.  The new package names are "profiles-registry", "motion-controller", and
"viewer".  There is no functional change in this PR, though folders and npm scripts have been modified.  Once this has been merged, we're unblocked on doing a bunch of important next work items.


NellWaliczek:
Oh! And because this context is relevant~!  The next thing I need to do in this repo is split the profiles-registry package into two packages.  

One package will stay the profiles-registry package and will contain the registry stuff such as the ids to be reported for each device, the list of fallbacks, and the mapping of buttons/axes array entries into to logical components.  This will include a refactoring of the schema currently in the profiles-registry package to remove all parts referring to asset node names.

The second package will be called "assets" and will contain the 3D assets and the node names portions of the JSON files to allow for animations.

The other packages (viewer and motion-controllers) will be updated accordingly.

NellWaliczek:
Related to #70, #71, #76, #54, #26, and #8

TrevorFSmith:
Yep, this looks like a good way to separate concerns and allow authors and framework-writers to pull only the bits they need.

PR #78, opened 2019-09-07
Splits profiles-registry package into separate registry and assets packages

NellWaliczek:
(Apologies for dropping another massive PR.  This should be the last of the significant structural changes. For clarity, I'm expecting we'll still need to spend a bunch of time digging into the schema and content of the registry package, so I'm likely to request that feedback be filed as separate issues in order to land the structural changes)

The new registry package describes the data that must be exposed
consistently by all User Agents. For a specific piece of hardware, such
as an Oculus Touch controller, all User Agents must report the same
ordering of strings in XRInputSource.profiles. All User Agents must also
consistently map physical hardware components, such as thumbsticks, into
identical indexes within the Gamepad.buttons and Gamepad.axes arrays.
The schemas and JSON profile descriptions can be extended in the future
to unify additional XRInputSource behaviors and values.

The new assets package contains 3D asset files and JSON files with the
information necessary to animate the 3D assets in response to backing
hardware's state changes. The JSON files conform to a schema which
enumerates the relevant node names within the asset and describes visual
responses that should be applied to the 3D asset as Gamepad.buttons and
Gamepad.axes values change. The build step of the assets package merges
its profile JSON files with the matching profile JSON files in the
registry. These merged JSON files and assets can then be used directly
in the javascript library published by the motion-controllers package,
or by developers using their own asset loading and animation library.

The motion-controllers library remains 3D engine agnostic, but has been
signifcantly simplified to directly ingest the merged profile JSON files
published by the assets repository. Small changes were also made to the
viewer packages to adjust accordingly.


packages/assets/profiles/htc/htc-vive.json
```diff
@@ -0,0 +1,85 @@
...
+                            "states" : ["default", "touched", "pressed"]
+                        },
+                        {
+                            "rootNodeName": "TOUCHPAD_TOUCH_Y",
+                            "source" : "yAxis",
+                            "states" : ["default", "touched", "pressed"]
+                        }
+                    ]
+                },
+                "menu": {
```
jacobcdewitt: According to the WebXR Gamepads module [spec](https://immersive-web.github.io/webxr-gamepads-module/#xr-standard-gamepad-mapping), "Buttons reserved by the UA or platform MUST NOT be exposed." So, I don't think we should have this entry for a menu button unless it has something to do with legacy WebVR compatibility.

packages/assets/profiles/microsoft/microsoft-045e-065d.json
```diff
@@ -0,0 +1,112 @@
...
+                            "states" : ["default", "touched", "pressed"]
+                        },
+                        {
+                            "rootNodeName": "THUMBSTICK_Y",
+                            "source" : "yAxis",
+                            "states" : ["default", "touched", "pressed"]
+                        }
+                    ]
+                },
+                "menu": {
```
jacobcdewitt: See my other comment about exposing menu buttons.

packages/assets/profiles/microsoft/microsoft-045e-065d.json
```diff
@@ -0,0 +1,112 @@
+{
+    "profileId" : "microsoft-045e-065d",
```
jacobcdewitt: Why not use "samsung-odyssey" instead of the VID/PID? That's more human-readable and the convention in this repo seems to be that we should use readable names.

packages/assets/profiles/htc/htc-vive.json
thetuvix: Agreed for system buttons like Home.  However, Menu is generally available for native apps to use.

Are we agreed on which button a UA may use within each profile for its own UI?  If not, UAs could diverge and leave few buttons consistently available for page use.

Picking the single “UA menu” button for each profile seems like a critical thing for us all to align on in this registry.

packages/assets/profiles/microsoft/microsoft-045e-065d.json
thetuvix: Good question!

This naming directly flows from the WinMR Gamepad.id values used in the WebVR era as keys for render model CDNs - for example: https://github.com/BabylonJS/Babylon.js/tree/master/assets/meshes/controllers/microsoft/045E-065D

The key question now is whether

thetuvix: OpenXR as a layer underneath WebXR is moving away from directly exposing things like VID/PID as output, instead having apps negotiate the set of controllers (OpenXR interaction profiles) that they already know about.

Perhaps it’s OK for the UA in that world to inherit the new controller from the registry only the next time the UA revs.  In the meantime, OpenXR underneath will tell the UA that the user is using whichever UA-known controller is most similar to the user’s true controller.

thetuvix: This also brings up a further possibility to drive cross-UA conformance for UAs built atop OpenXR.  Another package of JSON mappings could map each WebXR input profile to its corresponding OpenXR interaction profile, defining which button/axis index maps to which OpenXR `/user` path.  OpenXR-based UAs could then simply merge in the latest OpenXR mappings once per release to inherit conformant support for all the latest controllers.

packages/assets/profiles/samsung/samsung-gearvr.json
```diff
@@ -0,0 +1,66 @@
+{
+    "profileId" : "samsung-gearvr",
+    "assets" : {
+        "none" : {
+            "path" : "none.glb",
+            "rootNodeName" : "none-controller-node"
+        }
+    },
+    "layouts": {
+        "none": {
```
toji: This should define a "left-right-none" the same way that the oculus-go and google-daydream profiles do, because while it's not an inherently handed device the platform will assign it a hand and we'd expect that to be communicated by the API.

packages/motion-controllers/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Amazon
```
toji: Same as above.

packages/registry/profiles/microsoft/microsoft-045e-065d.json
```diff
@@ -0,0 +1,32 @@
+{
+    "profileId": "microsoft-045e-065d",
+    "fallbackProfileIds": [ "generic-trigger-grip-touchpad-thumbstick"],
```
toji: Shouldn't this have the "b" variant in the fallback list? That's like our canonical example of when a non-generic profile is useful.

packages/viewer/README.md
```diff
@@ -0,0 +1,27 @@
...
+> npm run test
+
+To build just this package invoke the following command from the root of the repository:
+> npm run build -- --scope @webxr-input-profiles/viewer
+
+### Filing a bug
+Fill in the steps for filing a bug https://github.com/immersive-web/webxr-input-profiles/issues/52
+
+### Filing a pull request
+Fill in the steps for filing a bug https://github.com/immersive-web/webxr-input-profiles/issues/52
```
toji: > steps for filing a **bug**

Copy/paste error?

packages/registry/profiles/samsung/samsung-gearvr.json
```diff
@@ -0,0 +1,28 @@
+{
+    "profileId": "samsung-gearvr",
+    "fallbackProfileIds": [ "generic-trigger-touchpad"],
+    "layouts" : {
+        "none" : {
```
toji: Same issue as above. This device will report a left or right handedness based on platform configuration.

packages/assets/profiles/htc/htc-vive.json
NellWaliczek: Since this doesn't appear to have agreement yet, I'll file an issue to discuss whether "reserved buttons" includes buttons the UA will be reserving and how to ensure that is consistent.

packages/assets/profiles/microsoft/microsoft-045e-065d.json
NellWaliczek: @thetuvix...so many words... o_O
Let me respond to each comment block individually:
* As these strings were the ones Microsoft chose to expose in the WebVR timeframe, I will be merging this pull request as-is unless you (as the hardware owner) specifically tell me to use different ones.
* As you suggest, when new OpenXR hardware comes out that a UA yet isn't aware of, the UA may be able to intuit which of the generic profiles to report
* Please file a separate issue to discuss tracking consistent OpenXR mappings if you'd like to track that specifically.

thetuvix: Sounds good - the discussion in this thread is orthogonal to your refactoring.

thetuvix: Filed 2 new issues here.

NellWaliczek: Thanks!

packages/registry/profiles/microsoft/microsoft-045e-065d.json
NellWaliczek: d'oh.  copy paste fail.

packages/viewer/README.md
NellWaliczek: yeeeeep

NellWaliczek:
I'm still digging on the license issue and hope to have an answer shortly.  Meanwhile, I've addressed the remainder of the comments.  I also realized that I'd totally broken the local-files version of the viewer, so I've fixed it.

packages/assets/profiles/htc/htc-vive.json
NellWaliczek: https://github.com/immersive-web/webxr-input-profiles/issues/82

PR #84, opened 2019-09-17
Add registry JSON validation to viewer

NellWaliczek:
The previous PR refactored asset profile validation to be usable in the viewer's local-profile mode.  This does the same for the registry profile validation.


packages/registry/gulpTasks/copySchemasTask.js
```diff
@@ -0,0 +1,34 @@
...
+    if (key !== mainSchemaKey) {
+      dependencies.push(input[key]);
+    }
+  });
+
+  const schemas = {
+    mainSchema,
+    dependencies
+  };
+  return Buffer.from(JSON.stringify(schemas, null, 2));
```
toji: Silly nit: I had to go searching for what the magic number `2` meant here. Any chance you could use something like `const indentationSpaces = 2` here and in the other `JSON.stringify()` calls to assist with readability?

NellWaliczek: I currently use this in 28 of places throughout the repo.  Is this a change you're hoping for across the board?

PR #87, opened 2019-09-17
Update Microsoft profile id strings and assets

### LINKED ISSUE #80: Should we use "samsung-odyssey" instead of the VID/PID?
@jacobcdewitt:
Why not use "samsung-odyssey" instead of the VID/PID? That's more human-readable and the convention in this repo seems to be that we should use readable names.

@thetuvix:
Good question!

This naming directly flows from the WinMR Gamepad.id values used in the WebVR era as keys for render model CDNs - for example: https://github.com/BabylonJS/Babylon.js/tree/master/assets/meshes/controllers/microsoft/045E-065D  This was done because the WinMR native APIs exposes VID/PID for each controller, and so a UA then didn’t need to be updated when a new WinMR controller came out.

The key question now is whether this still makes sense given the broader profile system, where an app could make use of a more generic profile before UAs update.

_Originally posted by @thetuvix in https://github.com/immersive-web/webxr-input-profiles/pull/78_

NellWaliczek:
This fixes #80


NellWaliczek:
@thetuvix @jacobcdewitt Does this look ok to you?

thetuvix:
Thanks!

Should the `samsung-odyssey` profiles and assets be in the `samsung` folders?

jacobcdewitt:
LGTM

PR #101, opened 2019-10-10
Clarify naming of "source" properties

### LINKED ISSUE #86: The term "source" is overloaded across the schemas
The terms "component" is used consistently across WebXR and this repo, while the term "source" is quite overloaded.

**Component**
* [`layouts.*.components`](https://github.com/immersive-web/webxr-input-profiles/blob/master/packages/registry/README.md#components): The parts of an input source.

**Source**
* `XRInputSource`: An entire controller.
* [`visualResponses.source`](https://github.com/immersive-web/webxr-input-profiles/tree/master/packages/assets#visual-responses): The kind of a source's component's state that affects a visual response.
* [`selectSource`](https://github.com/immersive-web/webxr-input-profiles/blob/master/packages/registry/schemas/layout.schema.json): The ID of the component used for `onselect` events.

This ambiguity has continued to confuse me as I dig through the profiles, trying to keep straight exactly how the various entities connect.

**Strawman:**
For `selectSource`, we could rename it to `selectComponentId`.

For `visualResponses.source`, my instinct was to rename it to `property`, but that is used by `"property": "visibility"`.  We could rename it to something like `data` or `field` instead.

NellWaliczek:
This PR attempts to address the naming concerns raised in #86 

Registry property "selectSource" is now "selectComponentId" and VisualResponse property "source" is now "dataSource".  Happy to use different names if folks still think these changes are unclear.

/fixes #86

### DIFF
--- a/packages/assets/README.md
+++ b/packages/assets/README.md
@@ -115,14 +115,14 @@ For example
 ### Visual responses
 The visual representation of a motion controller in a VR must respond to reflect its physical state in the real-world.  For example, when a physical thumbstick is moved to the left, the virtual thumbstick should also move to the left.  The `visualResponses` array contains descriptions of all visual changes that can occur when a controller part is interacted with.
 
-Each element in this array must contain a `rootNodeName` property which references the node containing the rest of the nodes needed for the visualization. It must also contain a `source` property set to one of four values: `button`, `xAxis`, `yAxis`, or `state`.  This indicates which component property will be used to drive the visualization.  Lastly, the element must contains a `states` array which indicates the component states for which the visualization will apply.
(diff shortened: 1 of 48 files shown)


NellWaliczek:
Added the missing markdown changes.  But also, @thetuvix expressed he'd still like to see the word "source" not be used in "dataSource" so I've renamed it "componentProperty".

thetuvix:
Looks good to me!

thetuvix:
Note that this passes my test when something is hard to name of looking at how you'd described it in English when you weren't thinking about names :smile:

From https://github.com/immersive-web/webxr-input-profiles/tree/master/packages/assets#visual-responses:
> This indicates which component property will be used to drive the visualization.

PR #102, opened 2019-10-10
Define and enforce fallbackProfileIds rules

### LINKED ISSUE #99: Figure out and document stability of fallback chains
See also https://github.com/immersive-web/webxr-input-profiles/issues/98

I and @toji discussed this a bit yesterday, and we're interested in hearing what others think. It seems to us like we should have no stability constraints on fallback chains: we're free to update them however we wish, provided there is always a generic fallback. Perhaps it could be a bit more nuanced (e.g. "the generic fallback will never change"), but it feels like there should be a policy here, and we should document it.

If we do this, issues like #98 become less pressing.

The reasoning behind not having stability requirements is that if you're _relying_ on a certain fallback chain you can just as easily code for the specific device, the fallback chain basically only exists for unknown devices to be compatible with code written for similar devices.

NellWaliczek:
/fixes #99 

Clarifies the README.md text to match the rules in #99 and adds a check in the validation step to ensure that the last fallback is a generic.

### DIFF
--- a/packages/registry/README.md
+++ b/packages/registry/README.md
@@ -55,7 +55,7 @@ Fill in the steps for filing a pull request https://github.com/immersive-web/web
 ## Schema
 
 ### Profile id and fallback profile ids
-Profiles are required to have a `profileId` that uniquely identifies the profile and must conform to the format specified by the [WebXR Device API](http://www.w3.org/tr/webxr). A `profileId` cannot be changed once added to the registry; if changes are necessary a new `profileId` must be defined. Profiles are also required to have a `fallbackProfileIds` array containing profile ids defined in the registry that are considered acceptable fallback substitutes for profile. Profiles with the `generic` prefix may have this array empty.  All other profiles must have at least one fallback profile id, and the the last entry in the array must have the `generic` vendor prefix. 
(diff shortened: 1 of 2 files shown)


packages/registry/README.md
```diff
@@ -57,13 +57,13 @@ Fill in the steps for filing a pull request https://github.com/immersive-web/web
...
 
 ### Profile id and fallback profile ids
-Profiles are required to have a `profileId` that uniquely identifies the profile and must conform to the format specified by the [WebXR Device API](http://www.w3.org/tr/webxr). Profiles are also required to have a `fallbackProfileIds` array containing profile ids that are considered acceptable substitutes for profile. This array may be empty, but, if it is not, the ids it contains must match another existing profile detailed in this repository. 
+Profiles are required to have a `profileId` that uniquely identifies the profile and must conform to the format specified by the [WebXR Device API](http://www.w3.org/tr/webxr). A `profileId` cannot be changed once added to the registry; if changes are necessary a new `profileId` must be defined. Profiles are also required to have a `fallbackProfileIds` array containing profile ids defined in the registry that are considered acceptable fallback substitutes for profile. Profiles with the `generic` prefix may have this array empty.  All other profiles must have at least one fallback profile id, and the the last entry in the array must have the `generic` vendor prefix.
```
toji: > Profiles with the `generic` prefix may have this array empty.

This reads a bit awkwardly. Maybe "... may **leave** this array empty"? Your call.

Manishearth: To properly address #99 can we also mention the stability rules? I.e. "the contents of the fallback list may change, though in the most part changes will likely be in the form of additions"

Issue #92, opened 2019-09-20
Enum values should use "hyphen-case" rather than "camelCase"


thetuvix:
There were a few enum values I saw that were using `"camelCase"` rather than `"hyphen-case"`.

Here's the ones I noticed in the registry:
* `"faceButton"`
* `"xAxis"`/`"yAxis"`

NellWaliczek:
So "faceButton" isn't an enum value, but the other two are.  Do we want to restrict component names to follow the same formatting rules as profile ids?

thetuvix:
We are otherwise following that pattern for component names - it seems reasonable to make it a rule.

PR #104, opened 2019-10-15
Normalizes schema key formats

NellWaliczek:
* Changes xAxis and yAxis to be x-axis and y-axis in schema
* Requires component ids to follow a lowercase, dash separated format
* Adds build check that xr-standard componenents are of the correct type
* Fixes bugs in registry build time checks
* Updates all profiles to match the changes
* Fixed a few bugs in the profile viewer
* Cleaned up the way gamepad control elements are generated in the viewer

/fixes https://github.com/immersive-web/webxr-input-profiles/issues/92

### DIFF
--- a/packages/assets/schemas/visualResponses.schema.json
+++ b/packages/assets/schemas/visualResponses.schema.json
@@ -18,7 +18,7 @@
             "componentProperty": {
                 "description": "The property name driving this visualization",
                 "type": "string",
-                "enum": ["xAxis", "yAxis", "button", "state"]
+                "enum": ["x-axis", "y-axis", "button", "state"]
             },
             "states": {
                 "description": "The states the visualization applies to",
--- a/packages/assets/profiles/generic/generic-button.json
+++ b/packages/assets/profiles/generic/generic-button.json
@@ -17,7 +17,7 @@
     "layouts" : {
         "left-right-none": {
             "components": {
-                "faceButton": {
+                "face-button": {
                     "rootNodeName" : "button-node",
                     "labelAnchorNodeName" : "button-label-transform",
                     "visualResponses" : [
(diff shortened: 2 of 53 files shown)


packages/assets/schemas/visualResponses.schema.json
```diff
@@ -18,7 +18,7 @@
             "componentProperty": {
                 "description": "The property name driving this visualization",
                 "type": "string",
-                "enum": ["xAxis", "yAxis", "button", "state"]
+                "enum": ["x-axis", "y-axis", "button", "state"]
```
thetuvix: With the new name `componentProperty`, `"x-axis"` and `"y-axis"` both make sense to me - it's easy to think of a physical thumbstick component as being composed of an `"x-axis"` property and a `"y-axis"` property.

However, a physical A button component is not itself composed of a `"button"` property or a `"state"` property.  It's also not jumping out at me when to use `"button"` and when to use `"state"`. (is it float value vs. bool value?)

Could both `"button"` and `"state"` be replaced with just `"value"`?  Then all three would feel like properties that a given component can have.

NellWaliczek: I agree that `"button"` now seems a bit strange.  However it isn't the same as `"state"`.  `"state"` is the combined opinion of whether a component is pressed, touched, or not being interacted with.  Specifically, when an axis value is greater than the threshold of the deadzone, the component is considered "touched" whether or not it has an associated `GamepadButton` with the `touched` attribute set to `"true"`.

That said, I completely agree that `"button"` is confusing, and I'm absolutely open to renaming it.  `"value"` seems like it'd be a bit confusing since `x-axis` and `y-axis` are also values.... Any suggestions?

Either way, i'm inclined to do that in a different PR to get this one landed.  Objections?

thetuvix: Sounds good - let’s carve this off as a new issue.

PR #110, opened 2019-12-03
Condense oculus-quest and oculus-touch-s profiles

### LINKED ISSUE #109: Oculus Touch S / Oculus Quest profiles redundancy?
**oculus-quest** **oculus-touch-s** 
These two profiles are identical and this is because it is the same controllers pair, just used with different headset. Everything is going to be identical for them - mappings, renderings, fallbacks, etc. Why do we need both, I think we can actually leave only one of them - **oculus-touch-s**.

toji:
As per the discussion in #109, these will now both be oculus-touch-v2,
since they are identical controllers shared between two headsets.

/fixes #109


Artyom17:
Looks good!

Issue #89, opened 2019-09-17
Fixed convention about glTF node names?


thetuvix:
As I wrote up how #88 could simplify node references, I wanted to confirm that we really do need the indirection of allowing profiles to target model-specific node names vs. simply defining well-known names for these glTF nodes.

For example, here is a current `component` definition for an `"xr-standard-thumbstick"` component:
```json
"xr-standard-thumbstick" : {
    "rootNodeName" : "THUMBSTICK_PRESS",
    "labelAnchorNodeName" : "thumbstick-label",
    "visualResponses": [
        {
            "rootNodeName": "THUMBSTICK_PRESS",
            "source" : "state",
            "states" : ["pressed"],
            "minNodeName": "UNPRESSED",
            "maxNodeName": "PRESSED"
        },
        {
            "rootNodeName": "THUMBSTICK_X",
            "source" : "xAxis",
            "states" : ["default", "touched", "pressed"]
        },
        {
            "rootNodeName": "THUMBSTICK_Y",
            "source" : "yAxis",
            "states" : ["default", "touched", "pressed"]
        }
    ]
},
```

However, imagine that the rendering library simply observes the model to have the following nodes:
* _(any name)_
  * `xr-standard-thumbstick-pressed-min`
  * `xr-standard-thumbstick-pressed-max`
  * `xr-standard-thumbstick-pressed-value`
    * _(any name)_
      * `xr-standard-thumbstick-xaxis-min`
      * `xr-standard-thumbstick-xaxis-max`
      * `xr-standard-thumbstick-xaxis-value`
        * _(any name)_
          * `xr-standard-thumbstick-yaxis-min`
          * `xr-standard-thumbstick-yaxis-max`
          * `xr-standard-thumbstick-yaxis-value`
            * _(any name)_
  * `xr-standard-thumbstick-label`

At that point, the rendering library can know purely by convention how to articulate the `"xr-standard-thumbstick"` component of that input source and find its label node.

Profiles could still define custom components such as `"a-button"`, which would then follow the same `pressed` naming convention:
* _(any name)_
  * `a-button-pressed-min`
  * `a-button-pressed-max`
  * `a-button-pressed-value`

Do you see ways in which this kind of convention could limit how model authors lay out their models?  For example, have you encountered cases where a thumbstick's `"default"`, `"touched"` and `"pressed"` states won't all articulate the same glTF nodes?

[...]

Issue #88, opened 2019-09-17
Referenced node names should be unique within a model


thetuvix:
One piece of feedback we've gotten around Windows MR's glTF-based controller models is that it's awkward to deal with ambiguous node names like `"UNPRESSED"` that only get their context by the name of their ancestor node (e.g. an `"UNPRESSED"` node that is a child of a `"GRASP"` node).

This is for two reasons:
* Code for rendering/articulating the controller model can more easily just look up the one node in the model with a given name, rather than finding an ancestor node with one name and then searching through its descendant nodes to find another name.
* glTF tools for authoring/transforming models in the first place can often expect that node names are globally unique within the model.

It may simplify things to establish a stronger rule here than we'd established in the native WinMR render model APIs, and just require that node names referenced in the assets profile show up only once in a given glTF file.  For the example above, the `"UNPRESSED"` node under `"GRASP"` would be renamed to `"GRASP_UNPRESSED"`.  This could also remove the need for profiles to specify the `rootNodeName` for each component and visual response, unless we want those for other reasons.

NellWaliczek:
Fixed by #118

PR #118, opened 2020-01-03
Remove boilerplate definitions from asset profiles

NellWaliczek:
This change is intended to address issues #88 and #89 by defining default naming conventions for all nodes necessary to utilize and associated asset.  These default names are automatically populated into the merged profile during the package build step unless overridden by asset's JSON file.


NellWaliczek:
Was thinking about this over the weekend, and realized we probably can get away with not having profile.json in the profiles folders if the asset will follow all the default behaviors.  I'm inclined to leave it this way in the current PR since the generic assets aren't in the folder structure yet, and we can make that change when the generic assets are ready to be added.

PR #122, opened 2020-01-07
Rename all asset node names to use '_' instead of '-' to be compatible with Maya

NellWaliczek:
Rename all asset node names to use '_' instead of '-' to be compatible with Maya


toji:
This change will also need to be applied to the table added in #121, right?

packages/assets/src/buildAssetProfile.js
```diff
@@ -183,14 +183,17 @@ function buildAssetProfile(assetInfo, expandedRegistryProfile) {
...
-        component.touchPointNodeName = `${componentId}-axes-touched-value`;
+        component.touchPointNodeName = `${componentId}_axes_touched_value`;
       }
     });
   });
 
   // Override any properties enumerated in the asset description file
   applyAssetOverrides(profile, assetInfo);
 
+  // Replace all the "-" in component and asset names with the Maya compatible "_"
```
toji: Is this the right approach, or should the components/assets be expected to use _ and any use of - is considered an error?

packages/assets/README.md
```diff
@@ -79,27 +79,27 @@ Or:
...
 
 ### Layouts
-Within each layout, there must be a `rootNodeName`, `assetPath`, and/or a `components` property. The `rootNodeName` is the top node in the 3D asset hierarchy representing the motion controller, and is set to `<profile id>-<handedness>` by default. The `assetPath` is the relative path to the asset for the layout, and is set to `<handedness>.glb` by default.  The `components` property is explained further in the [components](#components) section.
+Within each layout, there must be a `rootNodeName`, `assetPath`, and/or a `components` property. The `rootNodeName` describes the top node in the 3D asset hierarchy representing the motion controller. As Maya files cannot name nodes with "-" characters, the default value for this node name is `<profile id>_<handedness>` with all "-" in the profile id changed to "_". The `assetPath` is the relative path to the asset for the layout, and is set to `<handedness>.glb` by default.  The `components` property is explained further in the [components](#components) section.
```
toji: Make sure this actually renders right in GitHub. In the review preview I'm seeing it detect the "_" as the start of italics. But I also know there's differences between how this view detects the markdown and how the markdown actually renders. If it's being problematic probably just surround it in backticks instead: `_`

packages/assets/src/buildAssetProfile.js
NellWaliczek: Any use of "-" is an error.   Once i made the readme changes, this approach looked ooky though.  I've made a fix and as a result I've also changed the asset override schema to only accept "_"

packages/assets/README.md
NellWaliczek: I can't seem to find where you're seeing that issue, but I just added the backticks.  Can you double check that the error you're seeing is gone?

PR #126, opened 2020-01-09
Adding generic models

NellWaliczek:
Credit for all the beautiful models goes to Amaru Zeas! https://www.amaruzeas3d.com/


NellWaliczek:
I know it's generally frowned on to merge your own PRs... Breaking the rule so these beautiful assets are merged before I start my leave of absence.

AdaRoseCannon:
The models look so good!!

RaananW:
The models are amazing :-)

Not sure if this is the right place to ask, but i noticed that the x-axis of all generic models react differently than the mixed-reality model (which is, ATM, my point of reference). Positive X should go to the right, but in all generic models it goes to the left. 

Is that the correct behavior?

PR #128, opened 2020-01-10
Refactor viewer to enable viewing controllers in vr

NellWaliczek:
This refactor separates the boilerplate three.js code from the part that is specific to using the webxr-input-profiles libraries.  It also unifies the profile selectors so that any local profile is added to the dropdown and will be used if the actual VR controller would be matched to it.

(and with this, I'm out!)


packages/viewer/src/controllerModel.js
```diff
@@ -0,0 +1,159 @@
...
+      });
+    });
+  }
+
+  /**
+   * Walks the model's tree to find the nodes needed to animate the components and
+   * saves them for use in the frame loop
+   */
+  findNodes() {
+    this.nodes = {};
```
toji: You're setting `this.nodes` to a dictionary here, but initializing it to an empty array in the constructor.

packages/viewer/src/modelViewer.js
```diff
@@ -1,98 +1,64 @@
...
+  vrController.addEventListener('connected', async (event) => {
+    const controllerModel = new ControllerModel();
+    vrController.add(controllerModel);
+
+    const motionController = await profileSelector.createMotionController(event.data);
+    await controllerModel.initialize(motionController);
   });
+
+  vrController.addEventListener('disconnected', () => {
+    vrController.remove(vrController.children[0]);
```
toji: Answering my own question from above: Yeah, this is probably better than a `clear()` method in this context.

packages/viewer/src/controllerModel.js
```diff
@@ -0,0 +1,159 @@
...
+      );
+    }));
+
+    this.rootNode = this.asset.scene;
+    this.addTouchDots();
+    this.findNodes();
+    this.add(this.rootNode);
+    this.loaded = true;
+  }
+
```
toji: Not a problem, just curious: Did you determine this didn't need a `clear()` any more?

PR #133, opened 2020-01-13
Added Oculus Go controller assets

toji:
Don't want to land this until I've given it a try with actual hardware, but at least it all looks good in the viewer! (Aside from a lighting artifact that I also saw with the Daydream controller and now suspect to be a viewer bug instead of an asset issue.)

Mesh pulled from Aframe's asset repository (https://github.com/aframevr/assets/tree/master/controllers/oculus/go) and rigged with the appropriate animation nodes in blender by myself

On a side note: I realize that it's a bit weird to start my contributions to this repo with controllers for devices that have had their respective manufacturers signal that they're not long for this world (Daydream and the Go) but I figure that if I'm gonna learn the ropes on rigging these assets up my first and worst attempts ought to be on the assets that are likely to die off most quickly, right? 😜 Plus these are easier to deal with as they lack some of the more complex articulation of other devices.


toji:
Tested on real hardware tonight. Trigger/touchpad work as expected. Angle of the device seems slightly askew from what I would expect, despite being authored to spec. After some investigation it seems like it may just be the poses the browser is reporting. In any case, given that it's a 3DoF device it's not as noticable as it would be for a 6DoF controller. Can tweak later if needed.

PR #141, opened 2020-01-15
Added assets for the first gen Oculus Touch

toji:
Static meshes pulled from https://github.com/aframevr/assets/tree/master/controllers/oculus


Artyom17:
Thanks, @toji for all the assets you are adding!

toji:
My pleasure! Trying to get something in place for every profile we have in the registry (where it makes sense, anyway.)

PR #142, opened 2020-01-16
Created a tutorial for asset preparation with Blender

toji:
Dumping some of my newly acquired knowledge into a tutorial for future contributors!

Special request for @AdaRoseCannon to take a peek from a devrel point of view (Though this isn't a process most developers will have to go through.)


AdaRoseCannon:
I'll take a look :)

packages/assets/tutorial/README.md
```diff
@@ -0,0 +1,239 @@
...
+
+For a simple button component the default node names are `<buttonName>_pressed_value`, `<buttonName>_pressed_min`, and `<buttonName>_pressed_max` where any dashes (`-`) in the buttonName are replaced with underscores (`_`). To make things easier, the profile viewer page from earlier will also tell you some of the node names it's unable to find. In my case the Oculus Touch's "A" will need a `a_button_pressed_value`, `a_button_pressed_min`, `a_button_pressed_max`.
+
+I typically find it easiest to create the value node first and position it. While it's not required, it's helpful to not only align the value node with the button's center, but also rotate it so that the axis points stright up out of the button.
+
+![Aligning a node with a button](./images/tutorial-15.png)
+![Adjusting the node orientation](./images/tutorial-16.png)
+
+Once positioned, right click and copy the node in the outliner, then paste two copies and rename them to the "min" and "max" variants respectively, which gives us all three nodes at the same place. In the outliner move those nodes under the "root" node, then drag the mesh for the corresponding component and place it under the "max" node.
```
AdaRoseCannon: I think you should tell the user to delete the mesh from the copy and pasted min and max node, and that you will temporarily put the orignal geometry in it for testing.

Or perhaps position it with the geometry, then delete the geometry?

AdaRoseCannon:
This is great, just one thing to make it clear that it is empty meshes a but earlier but aside from that I love it.

packages/assets/tutorial/README.md
toji: Thanks, I've tweaked the language to try and make the operations performed on the mesh a bit clearer.

PR #144, opened 2020-01-16
Add generic-touchscreen input source profile

bialpio:
It's based on generic-touchpad profile - we could reuse it as-is, but it's
probably better to have a separate profile in this case as touchscreen and
touchpad are conceptually different.

The touchscreen could expose a gamepad object - this should make it easier
for the apps to compute screen space coordinates of the touched point, or
use the values as-is if that's sufficient for their use case.

Closes: immersive-web/hit-test#69.

@toji - please take a look!

### DIFF
--- a/packages/registry/profiles/generic/generic-touchscreen.json
+++ b/packages/registry/profiles/generic/generic-touchscreen.json
@@ -1,0 +1,24 @@
+{
+    "profileId" : "generic-touchscreen",
+    "fallbackProfileIds": [],
+    "layouts" : {
+        "none" : {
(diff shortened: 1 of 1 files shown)


toji:
I'm not entirely sure how I feel about the touchscreen advertising itself as a touchpad, because the interaction model is slightly different. (Touch is the select action for a touchscreen while press is the select action for a touchpad.)

However, I don't think the profile schema allows you to do anything more accurate at the moment, and I'd prefer to have this profile name in the registry ASAP, so I'm going to merge as-is and we can worry about a better internal representation later if needed. Thank you!

PR #147, opened 2020-01-17
Added assets for the HTC Vive Cosmos

toji:
The materials on these controllers could use some work, as the source assets that I had didn't export to glTF particularly cleanly so I had to strip some info out (like the roughness). Mesh is pretty good, though.


toji:
Okay, they have better materials now.

Manishearth:
This is failing CI

PR #148, opened 2020-01-20
Updated fetchProfile to allow a defaultProfile to be passed

### LINKED ISSUE #14: Add behavior for unknown gamepads
"unknown" gamepad ids may appear if the hardware is unidentifyable by the user agent or if the user agent has provided users a way to mask their gamepad id for privacy reasons.  In this situation, a mapping and asset should be present to fall back to.  This same fallback should be used if the gamepad id does not have a mapping in the repo.

toji:
/fixes #14 

This is the profile that'll load if none of the profile in the inputSource are matched. Debated about whether or not to pick a default but this feels like the least surprising way forward, as otherwise we might create a lot of false-positive controller "hits" on devices that don't actually match what's loaded at all. Doing it this way allows the developer to be intentional about their desire to fallback and select a profile that matches their needs most closely.


packages/motion-controllers/README.md
```diff
@@ -88,7 +88,7 @@ function onAnimationFrameCallback(xrFrame) {
 ```
 
 ## Components
-Once a `MotionController` has been created, developers can access interact with its components such as thumbsticks, touchpads, triggers, squeezes, and buttons. These components expose their current values through the `Components.values` object.  The `values.state` key will always be present and describes the overall state of the component as being `pressed`, `touched`, or `default`.  In addition to `values.state`, components may optionally also have a `values.button`, `values.xAxis`, or `values.yAxis`.  Each of values are populated slightly differently based on the underlying component type.
+ developers can access interact with its components such as thumbsticks, touchpads, triggers, squeezes, and buttons. These components expose their current values through the `Components.values` object.  The `values.state` key will always be present and describes the overall state of the component as being `pressed`, `touched`, or `default`.  In addition to `values.state`, components may optionally also have a `values.button`, `values.xAxis`, or `values.yAxis`.  Each of values are populated slightly differently based on the underlying component type.
```
Manishearth: The first clause might have been deleted here?

toji: Thanks for catching that! Fixed

Issue #155, opened 2020-01-24
Compressing assets textures and a possible asset pipeline optimization?


Giorat:
### What have I done?
I have done a rough compression of some of the biggest assets combining a simple decompression from glb to gltf with texture extraction, compression of textures and then went back again to glb.
I'm doing these tests in my fork: https://github.com/Giorat/webxr-input-profiles

I used:
- [gltf-pipeline](https://github.com/AnalyticalGraphicsInc/gltf-pipeline)
- [tinypng-cli](https://github.com/websperts/tinypng-cli)

All done with a simple bat file cause I'm using windows but should be similar to what you can do with a simple linux/mac script.

### Idea for an optimization pipeline
If it's needed I could also add in a folder you want a more complete node script to automatically compress new models for the future assets? This could be also useful if we want to produce automatically models with Draco compression or Basis texture easily.

### Side question always on models

I have seen that some of the assets have png textures and others have jpg. What format should be used on all of them? Using png could help to keep the quality fixed without artifacts due to compression.

Giorat:
Taken from @toji comment on my closed first PR.
"I tried reading up on tinypng, but I'm still fuzzy on a couple of things. It is a lossy compression algorithm, right? While that is probably OK for base color maps I'm concerned that it would introduce noticable artifacts into non-sRGB inputs, such as normal maps or metallic-roughness maps. Also, does repeatedly running tinypng on an image cause stacking artifacts? And just to verify, the compressed files are still just standard PNGs, right?

To answer a question from the related issue: I would be totally fine with changing any JPG textures to PNG as long as it doesn't significantly increase file size. Many of the textures have large sections of flat or similar colors, so I suspect that we should be OK on that front with the tinypng compression."

Will post more options as comment here to discuss them before doing any other new PR on this issue.

AdaRoseCannon:
Putting the compressing on the texture will help with download size but it will be loaded uncompressed on the end user's hardware.

Once Basis support is in GLTF it maybe better to use that format to get the compression benefits on both download and in graphics memory.

An additional form of compression which can be applied would be Draco compression for the Geometry. Which is something I think is already supported in GLTF but I haven't tried it myself yet.

Giorat:
> Putting the compressing on the texture will help with download size but it will be loaded uncompressed on the end user's hardware.

In some cases this won't happen if you simply decrease textures colors.
But it will change for sure with Draco that will require some decompression time.

toji:
[...]

PR #156, opened 2020-01-24
Major compression of assets textures

Giorat:
**MAJOR changes**
I compressed all models textures as I already explained in [this issue: Compressing assets.](https://github.com/immersive-web/webxr-input-profiles/issues/155)

**Minor changes**
- "http-server" version upgrade to [fix a problem on Windows machines](https://github.com/http-party/http-server/issues/525#issuecomment-574152956)
- dist folder has been added to ".gitignore" to prevent commits of the dist"s" folder.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2777,6 +2777,16 @@
         }
       }
     },
+    "@webxr-input-profiles/assets": {
+      "version": "0.1.0",
+      "resolved": "https://registry.npmjs.org/@webxr-input-profiles/assets/-/assets-0.1.0.tgz",
+      "integrity": "sha512-FV+rOTZ3shZvryLDYCK27Rk3M1yP9clhr29f/xtselBI3HtiVWllRr91OzENoLFtL5iLXFyyge3PTyPFHp9aww=="
+    },
+    "@webxr-input-profiles/motion-controllers": {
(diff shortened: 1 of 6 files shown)


toji:
Thanks for the contribution! This seems like a valuable change, given that we expect these meshes to be downloaded a lot. I do have some questions/concerns, though:

I don't believe that we can add the `dist/` folders to .gitignore, because it would prevent our gh-pages deployment step from picking them up, and we need that for the viewer.

I tried reading up on tinypng, but I'm still fuzzy on a couple of things. It is a lossy compression algorithm, right? While that is probably OK for base color maps I'm concerned that it would introduce noticable artifacts into non-sRGB inputs, such as normal maps or metallic-roughness maps. Also, does repeatedly running tinypng on an image cause stacking artifacts? And just to verify, the compressed files are still just standard PNGs, right?

To answer a question from the related issue: I would be totally fine with changing any JPG textures to PNG as long as it doesn't significantly increase file size. Many of the textures have large sections of flat or similar colors, so I suspect that we should be OK on that front with the tinypng compression.

As for the http-server change, I'd prefer that be separated out into a separate PR, just to keep this one focused on the asset change.

Finally, I'm afraid that a change I made this morning has introduced a conflict in this path

PR #157, opened 2020-01-25
Update http-server to 0.12.1 to fix windows problem

Giorat:
Update http-server to 0.12.1 to fix windows problem: https://github.com/http-party/http-server/issues/525#issuecomment-574152956


toji:
Awesome, thanks!

PR #158, opened 2020-01-28
Updated input profiles to match current OpenXR Chromium implementation

fordacious:
Updated the input profiles to match what is currently implemented in the OpenXR implementation of WebXR for Chromium, which has input bindings for all PC VR controllers.

Each input profile has a menu button reserved by the browser.


Manishearth:
Thanks!

PR #162, opened 2020-02-24
Add Magic Leap Asset

atarng-magicleap:
Add Magic Leap Control assets to viewer.


toji:
/agenda to talk about whether we should be delivering assets for AR-only devices to users.

I'd love to have this asset in the repo and available for users if they have a need for it, but I'd like to have some further discussions on whether or not this is something that we should actively be pushing to users of the library, since it seems like you probably don't want to actually be rendering this asset on the intended device. If you did it would just end up rendering (probably imperfectly aligned) over the top of the real device.

PR #167, opened 2020-04-08
Add HTC Vive Focus Plus profile and model.

daoshengmu:
I would like to add HTC Vive Focus Plus profile and its model. I confirmed it work properly in Viewer and got acknowledge from HTC.


toji:
Excellent, thank you!

PR #169, opened 2020-05-28
Added Pico Interactive registry prefix.

david-pico:
We're working on WebXR support for our XR headsets and would like to reserve "pico" for our company prefix.


toji:
Great, thanks!

PR #181, opened 2020-09-25
Added hp mixed reality controller assets

fordacious:
@toji


toji:
Checked out locally and verified that the models load as expected and all the animation rigging is working properly. Excellent work, thanks!

PR #189, opened 2021-01-15
Added w3c.json for WG/note

himorin:


w3c.json
```diff
@@ -0,0 +1,5 @@
+ {
+    "group":      109735
+,   "contacts":   ["dontcallmedom", "himorin"]
+,   "repo-type":  "note"
```
dontcallmedom, line 4: is the intent really to publish this as a WG Note? My sense is that it's more of a CG item, and probably more in the "tool" or "other" repo-type.

himorin: changed to "tool"

w3c.json
```diff
@@ -0,0 +1,5 @@
+ {
+    "group":      109735
```
dontcallmedom, line 2:
```suggestion
    "group":      87846
```

dontcallmedom (approved):

PR #197, opened 2021-04-27
Add generic WebXR hand model

fe1ixz:
Adding universal WebXR hands model


cabanier:
@toji This model should work with all implementations of WebXR Hands.

toji:
Thank you! Much appreciated!

cabanier:
Thanks @toji ! Will this also update the CDN?

toji:
No, I have to do that myself. I wanted to do a quick verification before I pushed to the CDN

toji:
com oculus shellenv-20210427-163128

Verified that I can coerce Three.js into using these meshes successfully, so the basic litmus test of "are they usable" passes. :) And I'd definitely prefer to have Three using these rather than the FBX models it's using now, regardless of joint orientation issues.

There is still some residual weirdness that I'd like someone to look at. For one, applying the joint positions in Three requires them to be scaled by 100, and I don't know if that's an issue with Three, with the hand data coming out of the oculus browser, or with the meshes. I'll look into it. Second, there's a LOT of noisy joint data in the hand files that it seems like we could clean up without any ill effects? Maybe it's just a side effect of how Blender imports them, but it looks like we've got twice as many joints as we need and a lot of superfluous axes. It's not a deal breaker, but given that we expect/hope that these hands will be used as a default mesh by many WebXR apps it's worthwhile to try and reduce unnecessary bloat in the files.

Screenshot 2021-04-27 163923

Regardless, I'll try to get the CDN updated tonight and we can iterate on the rest of it.

toji:
Published. Version will be 1.0.6. Not sure how long it takes to start showing up on https://www.npmjs.com/package/@webxr-input-profiles/assets or the CDN.

cabanier:
awesome! Thank you so much.

toji:
Okay, looking into it some more this morning, the "scale by 100" thing is definitely a problem with the model. Looking at the mesh in blender I can see that all of the bones have a scale of `0.01`. If I select the entire model, then use `Object->Apply->Scale` (which bakes the currently applied scale into the vertices/bones/etc) then I can get rid of the multiplier in Three and it still works well. As a result I'm working on a cleaned up version of the hand assets that I'll post as a separate pull request and ask for the Oculus crew to review.

Maksims:
Looking into model internals, they have a lot of nodes, with names not consistent with WebXR APIs, making it harder to understand which are just anchor nodes, and which are bone nodes.
Are those nodes necessary, and what are their purpose? I believe it would be great to have the most simple model of a hand, with just WebXR Joint names as nodes, that other artists will have the ease to base their hands off.

And yes, "Armature" node is scaled `0.01`, that is why scaling 100 times is required.

https://playcanvas.com/viewer?load=https://cdn.jsdelivr.net/npm/@webxr-input-profiles/assets@1.0.6/dist/profiles/generic-hand/right.glb

fe1ixz:
> Okay, looking into it some more this morning, the "scale by 100" thing is definitely a problem with the model. Looking at the mesh in blender I can see that all of the bones have a scale of `0.01`. If I select the entire model, then use `Object->Apply->Scale` (which bakes the currently applied scale into the vertices/bones/etc) then I can get rid of the multiplier in Three and it still works well. As a result I'm working on a cleaned up version of the hand assets that I'll post as a separate pull request and ask for the Oculus crew to review.

Thank you for the great work! Just saw your post of the cleaned up model, I will try it out in experience shortly

toji:
> I believe it would be great to have the most simple model of a hand, with just WebXR Joint names as nodes, that other artists will have the ease to base their hands off.

Renaming the bones to match the WebXR joint names is a good idea. I'll incorporate that into #198

PR #198, opened 2021-04-28
Applied bone scale, renamed bones, and removed unused bones and nodes

toji:
This is a straightfoward cleanup and bug fix of the hand meshes added by @fe1ixz in #197.

I found that the bones had a scale of 0.01, which caused pages that wanted to use the meshes to need to scale their joint positions by 100, which isn't desirable. To correct it I opened the glTF files in Blender, used `Select->All` and `Object->Apply->Scale`. This normalized all the scales to 1.0 without affecting the appearance of the mesh or the position and weights of the bones.

Second, there were a lot of bones and nodes in the mesh that didn't correspond to any aspect of the WebXR hand API, weren't part of the skinning, and didn't seem well aligned with the mesh. You can see them in the image below:

Screenshot 2021-04-27 163923

I removed deleted these apparently unnecessary elements and confirmed that the mesh still responded appropriately when used in an environment like Three.js. Happily, this reduced the file size by 24Kb each!

Screenshot 2021-04-28 100736

Please let me know if you feel the above changes adversely affect the functionality of the hand meshes.


fe1ixz:
Screen Shot 2021-04-28 at 10 19 31 AM
Loaded up in blender and it looked like this, not sure if it is expected. Trying out in one of our hand experience now.

fe1ixz:
Tried it out with THREE.js' hand example, this renders normally, thank you for the clean up!

cabanier:
> Tried it out with THREE.js' hand example, this renders normally, thank you for the clean up!

Thanks @fe1ixz  and @toji

cabanier (approved):

toji:
As suggested by @Maksims in #197 I've also now updated the names of each bone in the file to match the WebXR enum names, then updated the Three.js example to use those as well. This reduced the amount of setup code significantly, and made understanding what it was doing quite a bit easier! Still renders correctly too:

https://user-images.githubusercontent.com/805273/116451105-edf2fd80-a810-11eb-87cd-e82e9e8b9bca.mp4

Renaming the bones means that anyone that was previously using the Oculus FBX mesh will have to refactor their app a little bit to make use of this asset, but I feel pretty strongly that it's a win in terms of clarity.

toji:
The Three.js changes needed to support this new asset are here, in case anyone is interested: https://github.com/mrdoob/three.js/pull/21743

PR #199, opened 2021-07-07
Add HTC Vive Focus 3 profile and model.

lawwong:
Hi, I would like to add profile and models for Focus 3 controllers.
I verified in the profile viewer and the model looks fine.


toji:
Thank you very much for this contribution! Apologies that I didn't see it before now.

In testing the new assets with the viewer, everything appears to be working great with one exception: The joystick axes seem to be inverted. If you could update the assets to reverse the joystick motion on both axes of both controllers, I'd happily merge these.

For reference: WebXR follows the Gamepad API's axis mapping, which is +X is to the right and +Y is down.

webxr-axes

jefmes:
I'm a new Focus 3 user and have been trying to figure out why I'm not getting any controllers or input in Mozilla Hubs, and I stumbled into this during my search. I have an issue open over on the Mozilla Hubs GitHub page - am I correct in thinking once this profile is properly added and I assume the latest additions will need to be merged into Firefox Reality, this may resolve the problem?   https://github.com/MozillaReality/FirefoxReality/issues/3945

lawwong:
@toji
Thanks for the reminding!
I have just uploaded the flipped asset, please help review again, thanks a lot.

toji:
Excellent! Everything is responding like I would expect now, thanks!

jefmes:
> Excellent! Everything is responding like I would expect now, thanks!

So...apologies for my ignorance on the next steps here, but what would still be required for this to start impacting downstream for the users? Once these are merged does it automatically/dynamically start working through the magic of the interwebz? :)  Or will all of the WebXR compatible platforms need to push out code changes to users in next versions of Chrome/Firefox/etc. I'm concerned a bit on the Focus 3 because of the lack of support behind Firefox Reality at the moment.

toji:
No worries! It's a fair question, and the answer depends on how you use the assets.

Some libraries like Three.js default to pulling their assets from a common CDN. We somewhat encourage using [jsdelivr](https://cdn.jsdelivr.net/npm/@webxr-input-profiles/assets@1.0/dist/profiles/) because the more places that use a common CDN the better the end user's caching will be. If a CDN is used you'll start "magically" seeing the new assets shortly after the package is published (which I just did).

WebXR apps can also choose to maintain and serve their own copy of the assets, in which case they won't pick up the new assets until they have manually rolled the changes into their codebase.

Finally, even once the assets are available they won't be picked up unless the browser you are using advertises the correct input profile for the device. On standalone headset's browsers this usually isn't a problem, since they can hardcode it to report the right profile, but if you are using your device with desktop Chrome via a cable or WiFi connection you'll probably see the wrong controllers simply because Chrome hasn't been updated to recognize the device.

jefmes:
Thanks so much Brandon, great explanation.

PR #203, opened 2022-01-04
Add PicoVR Neo3 controller models and profiles

song-fangzhen:
Pico(Bytedance) would like to contribute the controllers we use to the webxr-input-profiles repo. These models already got acknowledged by @picoxr.

- [x] PicoVR Neo3 controller model and profile


song-fangzhen:
@Manishearth @toji PTAL, thanks!

toji:
Loaded this locally and it looks great! Thank you for taking the time to rig the models and set up the profile.

song-fangzhen:
Hi, @toji.
May I ask when the assets on [CDN](https://cdn.jsdelivr.net/npm/@webxr-input-profiles/assets@1.0/dist/profiles/) will be updated \
 so that any browsers/websites that want to use the latest modes can make a test with them?

toji:
Thanks for the reminder. I will try to push that today.

toji:
Published: https://www.npmjs.com/package/@webxr-input-profiles/assets

PR #220, opened 2022-03-31
Update hand metacarpals to match OpenXR

cabanier:
I will not submit this PR until the Oculus browser switches to OpenXR


toji:
To answer the question posed on the merge but which is for some reason not showing up on this PR: No, simply merging won't push to the CDN, but I'll update it manually soon.

PR #225, opened 2022-10-08
Fix PicoVR Neo3 controller models

### LINKED ISSUE #223: pico-neo3 glb are broken on some viewers
**Profile id**
pico-neo3

**Describe the bug**
On some glTF viewers, the pico-neo3 model face is backward.

**To Reproduce**
When viewing the pico-neo3 profile models on the WebXR Input Profile Viewer https://immersive-web.github.io/webxr-input-profiles/packages/viewer/dist/index.html
The models faces are backwards
image
The same is with the Unity WebXR Input Profiles Loader https://de-panther.github.io/webxr-input-profiles-loader/Build/
image
Seems like there are issues with three.js r142 using https://gltf-viewer.donmccurdy.com/
image

**Expected behavior**
Khronos glTF-Sample-Viewer https://github.khronos.org/glTF-Sample-Viewer-Release/
image
Manage to load it as it should

The bug is in the viewers, but maybe the models should support older viewers for better reach.

De-Panther:
Resolve #223 

Following https://github.com/immersive-web/webxr-input-profiles/issues/223 and by @atteneder suggestion at https://github.com/atteneder/glTFast/issues/491
Set `alphaMode` `OPAQUE` and `doubleSided` `false`.

That fix the models for the viewers mentioned at #223

unpacked and packed using https://github.com/CesiumGS/gltf-pipeline


De-Panther:
@toji PTAL

toji:
Thank you! I just noticed this in the viewer yesterday, so I'm glad there's a fix.

De-Panther:
Thanks @toji .
and sorry, seems like after the pack it removed some of the nodes.
So now it needs another fix :/

PR #226, opened 2022-10-11
Update default hand position

cabanier:
The default position of the hands was not correct.
This has no effect if the hand model is used with joints but if an experience is using the hands as regular controller, they would be in the wrong location.

We rotated the hands to their proper position and adjusted the offset so they overlay the user's actual hands.
Thanks @intrepidolivia for helping me out!


Manishearth (approved):

hybridherbst:
@cabanier I think this PR may have accidentally removed all skinning information from the hands, see
- https://github.com/immersive-web/webxr-input-profiles/issues/233

cabanier:
@intrepidOlivia can you take a look?

PR #232, opened 2022-10-13
Fix PicoVR Neo3 controller models - Added removed nodes

De-Panther:
Following the bug from #225
Unpacked and repacked the PicoVR Neo3 models, this time keeping unused elements.
Made sure to test it on different viewers, and made sure the relevant nodes are in the models, while the models are `alphaMode` `OPAQUE` and `doubleSided` `false`.


De-Panther:
@toji PTAL

toji:
Thanks!

PR #240, opened 2022-12-08
Add PicoVR 4 controller models and profiles

song-fangzhen:
Pico(Bytedance) would like to contribute the controllers we use to the webxr-input-profiles repo. These models already got acknowledged by @picoxr.

- [x] PicoVR 4 controller model and profile


VBagmut:
@toji can you merge this please?

toji:
Thanks for the ping! I had somehow missed the PR previously. I'm going to do a quick due-diligence check on the models just to verify that the rigging is as expected then I'll merge them.

toji:
Glad I checked. The X axis of the thumbstick on the right controller is inverted. This means that when the X value is +1 the thumbstick is tilted all the way to the left, when it should be tilted all the way to the right. (This is a common issue when the rigging for the models is simply mirrored without adjustment during creation.)

@song-fangzhen: Can you please fix the inverted rigging of the right controller's thumbstick X axis? Then I'll happily merge them, as they're nice, high quality assets!

song-fangzhen:
Oops! Sorry for not realizing that before pushing the commit. Thanks for pointing out the issue and the fix has been committed, PTAL again!

toji:
Thanks for the speedy fix! Verified that the thumbstick is fixed and all other controls are still rigged properly.

PR #251, opened 2023-09-28
Add support for meta-quest-touch-plus controllers

felixtrz:
meta-quest-touch-plus is the profile name of the newly launched Meta Quest 3 controllers @toji


toji:
LGTM! Checked out locally and verified that all the control movements are as expected. Thanks for providing the mesh!

PR #252, opened 2023-10-18
motion-controllers: add annotations for tree-shaking

CodyJasonBennett:
Adds pure annotations for tree-shaking. `Object.freeze` is properly impure since it mutates its arguments, and top-level member expressions are pedantically seen as impure since getters can be called, so I wrapped it into an IIFE rather than destructured every level.

https://esbuild.github.io/api/#tree-shaking

`npx treeshake-cli@latest dist/motion-controllers.module.js`


toji (approved):
LGTM, thank you!

PR #253, opened 2023-10-18
motion-controllers: fix NPM config

### LINKED ISSUE #250: Outdated npm package
The content of the "package.json" inside the npm package (https://www.npmjs.com/package/@webxr-input-profiles/motion-controllers) doesn't match the content of the "package.json" in this repository.

In the npm package.json the "main" field points to the module variant, while the package is not declared of type module. In this repository the package.json correctly has the "main" field pointing to the cjs variant and the "module" field points to the module version.

Another note: The "module" field is not officially supported and some bundlers simply ignore it, making this library a CommonJs library for them.
Therefore, many libraries (e.g. three.js) have switched to declare their package as "type: module" and using the exports field:

```json
{
  "type": "module",
  "exports": {
    ".": {
      "import": "./build/three.module.js",
      "require": "./build/three.cjs"
    },
  }
}
```

CodyJasonBennett:
Fixes #250 and configures the `exports` field for Node ESM resolution. `module` is a Webpack convention and is non-standard.

As-is, the package resolves only for CJS which is not currently published on NPM due to the `files` field.

Might be worth adding https://publint.dev CLI to CI or before publish.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 node_modules
 npm-debug*.log*
 lerna*.log
+dist
--- a/packages/motion-controllers/package.json
+++ b/packages/motion-controllers/package.json
@@ -2,14 +2,20 @@
   "name": "@webxr-input-profiles/motion-controllers",
   "version": "1.0.0",
   "description": "",
-  "main": "dist/motion-controllers.js",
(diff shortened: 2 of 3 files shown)


toji (approved):
Thank you for this as well!

Issue #254, opened 2023-10-19
Profile Validator page and GitHub link are broken


hybridherbst:
**Describe the bug**
The Profile Validator linked from the Readme currently throws a fetch error:
https://immersive-web.github.io/webxr-input-profiles/packages/viewer/dist/index.html
![image](https://github.com/immersive-web/webxr-input-profiles/assets/2693840/28c01378-ed01-4abd-bb16-90690c07d95e)

**To Reproduce**
1. Open https://immersive-web.github.io/webxr-input-profiles/packages/viewer/dist/index.html
2. Note you can't choose profiles
3. Look at the console

**Desktop (please complete the following information):**
 - OS: Android, iOS, Windows
 - Browser Chrome, Safari

**Additional context**
The link on the repo is also broken right now:
https://immersive-web.github.io/webxr-input-profiles/
![image](https://github.com/immersive-web/webxr-input-profiles/assets/2693840/acaa234b-64fa-4d7f-897b-1a8ff98b9be0)

De-Panther:
Seems like this commit
https://github.com/immersive-web/webxr-input-profiles/commit/ec044381753a5d00154711b91cf6f50804df6f52
Replaced `motion-controllers.module.js` with `motion-controllers.js`

Searching for `motion-controllers.module.js`
https://github.com/search?q=repo%3Aimmersive-web%2Fwebxr-input-profiles+motion-controllers.module.js&type=code
Finds instances of this in files:
```
packages/viewer/rollup.config.js
packages/viewer/src/mocks/mockGamepad.js
packages/motion-controllers/README.md
packages/viewer/src/controllerModel.js
packages/viewer/src/profileSelector.js
```

De-Panther:
Site seems down due to `remote: Permission to immersive-web/webxr-input-profiles.git denied to github-actions[bot].`

himorin:
fixed by #258

PR #255, opened 2024-01-23
Fix motion-controllers module path

De-Panther:
Should resolve the viewer problems in issue #254

Following commit https://github.com/immersive-web/webxr-input-profiles/commit/ec044381753a5d00154711b91cf6f50804df6f52
I replaced all instances of `motion-controllers.module.js` with `motion-controllers.js`

I tested using:
```
npm install
npm run build
npm run dev
```
and it works locally


De-Panther:
cc @toji 
It looks like the checks fail is related to permissions.
`remote: Permission to immersive-web/webxr-input-profiles.git denied to github-actions[bot].`

AdaRoseCannon:
@himorin is fixing these permissions something we can do?

himorin:
> @himorin is fixing these permissions something we can do?

chatted with @toji during last f2f, and this PR could resolve one of JS error but not yet sure all will work.
for GHAction, it's system restriction that GitHub-Pages-deploy will fail for PR from external user of organization (lack of write permission). We've seek around several options but haven't reached to any way to go.

De-Panther:
Maybe someone from the org can clone and re-open the PR? (Won't help for future PRs, but for this one...)

De-Panther:
The error mentioned specific email, so maybe only this user can run the tests?

De-Panther:
I see that the fix was done on another PR, closing this.

PR #258, opened 2024-04-23
potential fix over js modules

himorin:
replaces #255 , incl. missing parts for fix in original PR.


himorin:
sigh,, check failed... will check later...

toji (approved):
LGTM. I'm inclined to merge this and see if it's able to deploy correctly. (After all, we're not at risk of being _more_ broken than it is currently, right?) I have a suspicion that the check script is faulty in this regard.

toji:
Let's see how that behaves...

toji:
Yeah, I see it working at https://immersive-web.github.io/webxr-input-profiles/packages/viewer/dist/index.html now as well. Thank you!

PR #259, opened 2024-06-07
add menu button to Quest/Oculus controllers

cabanier:
An upcoming version of the Quest browser will no longer exit WebXR when the menu button is clicked.
This will enable us to add support for this button to controllers and hands.


AdaRoseCannon:
Will there still be a system button to ext an XR session?

cabanier:
> Will there still be a system button to ext an XR session?

The Quest/Oculus button will still bring up trusted UI.

toji (approved):
LGTM, thanks for clarifying about the trusted UI!

PR #261, opened 2024-06-12
Add generic-fixed-hand registry entry

toji:
Only adds the registry entry, not an associated asset. That will be handled in a follow up.

Last call for bikeshedding! :laughing:

### DIFF
--- a/packages/registry/profiles/generic/generic-fixed-hand.json
+++ b/packages/registry/profiles/generic/generic-fixed-hand.json
@@ -1,6 +1,6 @@
 {
     "profileId" : "generic-fixed-hand",
-    "fallbackProfileIds": ["generic-hand", "generic-hand-select"],
+    "fallbackProfileIds": ["generic-hand-select", "generic-trigger"],
     "layouts" : {
         "left" : {
             "selectComponentId": "xr-standard-trigger",
--- a/packages/registry/profiles/generic/generic-fixed-hand.json
+++ b/packages/registry/profiles/generic/generic-fixed-hand.json
@@ -1,0 +1,32 @@
+{
+    "profileId" : "generic-fixed-hand",
+    "fallbackProfileIds": ["generic-hand", "generic-hand-select"],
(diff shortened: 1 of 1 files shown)


packages/registry/profiles/generic/generic-fixed-hand.json
```diff
@@ -0,0 +1,32 @@
+{
+    "profileId" : "generic-fixed-hand",
+    "fallbackProfileIds": ["generic-hand", "generic-hand-select"],
```
cabanier, line 3: should we fall back to `generic-trigger` instead?

toji: Hm... I could see an argument for it either way. Curious what others think, especially @AdaRoseCannon.

alcooper91: generic-trigger IIUC falls back to being basically a controller?

cabanier: yes. it would be strange to fall back to an animated hand that won't draw correctly

toji: What do you think about keeping `generic-hand-select` in the profiles array, since it's an accurate description of the input and doesn't have an associated asset?

cabanier: that sounds good to me

AdaRoseCannon: > What do you think about keeping `generic-hand-select` in the profiles array, since it's an accurate description of the input and doesn't have an associated asset?

Yeah that makes sense, to me.

toji: Thanks for the feedback! Updated the profile list to `['generic-fixed-hand', 'generic-hand-select', 'generic-trigger']`, which was my sense of the best "Most specific->Least Specific" progression.

PR #263, opened 2024-06-13
Add assets for generic-fixed-hand

toji:
Here's a pass at a `generic-fixed-hand` asset based on one that Ada emailed me. (I had to re-orient the mesh and I tweaked the finger positions). Screenshots are from the input profiles viewer to give you and idea of how it looks.

hand-angles

One unfortunate thing I discovered after playing around with the profile JSON for much longer than I probably should have is that it would be pretty difficult to make the little ball appear and disappear based on user input. There _is_ a mechanism for controlling visibility in the input profiles code, but at the moment there's a couple of quirks to it that effectively make it only work for controlling the visibility of the touch point on a touchpad, and making use of that would likely require the emulated hand controllers to surface inputs on `buttons[3]` on the gamepad. I could fix it in the code, but doing so would likely require additional updates to the libraries which are currently rendering the assets, and that defeats part of the purpose behind introducing this asset in the first place. 🤷 Anyway, I leave it up to the reviewers as to whether or not the ball serves any purpose if it's not visually responsive to users input. (We *can* reliably change it's position and orientation in response to input, so if anyone has thoughts on how that can be used here I'm all ears!)

Feedback is welcome on the hand pose as well! I tried for something where the index finger is mildly extended and the other fingers are progressively more retracted. Not quite a "pointing" gesture but something that you could use to poke things if needed while still feeling neutral. Ada's original asset was posed more like it was holding an invisible controller, with the index and thumb in more of a "about to pinch" position, and it may be that everyone feels that's a more natural pose for how this asset is likely to be used.

Screenshot 2024-06-13 132402


AdaRoseCannon:
I really like this pose a couple of thoughts regarding the animation:
* We could do an update to the library to make the ball appear and scale, and if so could we make it default to invisible for the old version?
* We could have 2-3 balls which move together/rotate in a cute way to indicate pinching but that might be too flavourful.

toji:
> We could do an update to the library to make the ball appear and scale, and if so could we make it default to invisible for the old version?

"Default to invisible" isn't really something the library does right now. You'd have to do something like set the scale to zero and then add scaling support.

> Instead we could have 2-3 balls which move together/rotate in a cute way to indicate pinching but that might be too flavourful.

I was thinking about something like this too! My thought was to maybe make the ball two-tone and flip it when pinching, but I like the colored ball cluster idea too!

One thing to consider is that with how OpenXR is structured at least, there's the potential that the pinch gestures could be reported in an analog way (ie: a value from 0 to 1 depending on how far apart the fingers are). This means that if we do any sort of translation/rotation it's possible for some future devices (I don't think Quest does this, and it wouldn't apply to Vision Pro) to animate that transition. Anything we do in this regard should probably keep that in mind. (Otherwise I would suggest we hide the ball in the palm of the hand and snap it out when you tap, but that would look weird in this scenario.)

cabanier (approved):

AdaRoseCannon:
> (I don't think Quest does this, and it wouldn't apply to Vision Pro)

I can't think of a situation where Vision Pro would use this model since we always have transient-pointer so don't need to use a fallback hand.

I think we should leave the decision to @cabanier as to whether we should remove the ball, hide it in the hand or do a position based animation.

toji:
SGTM.

For what it's worth, I tried out a two-tone ball and had it rotate 180 deg when you tap. I think that could work, as it at least gives some basic visual feedback that input is being recognized.

Screenshot 2024-06-13 142252
Screenshot 2024-06-13 142302

That said, if Rik doesn't think it's necessary I'll just drop the ball and merge.

cabanier:
yes, you can drop the ball

toji:
Ball removed! And with that, I'm going to merge.

PR #264, opened 2024-07-26
Add Magic Leap 2 controller assets

atarng-magicleap:
Adds Magica Leap 2 controller assets.


atarng-magicleap:
Squashed to just one Commit. after fixing new line issue.

idrisshah:
@toji can you please review this change?

toji:
Thank you for your contribution, and thanks for the ping! I was out of office when this was initially pushed and it fell off my radar.

The provided mesh looks very high quality, which is great! It does need some work to properly rig the buttons/triggers for use with this library so that the mesh visually responds to user inputs. Please see ["Preparing a WebXR input profile mesh"](https://github.com/immersive-web/webxr-input-profiles/blob/main/packages/assets/tutorial/README.md) for details on how to do that, and reach out to me if you have difficulty getting it set up properly.

I'm happy to help with the rigging if you need, but since I don't have one of the physical devices available I would only be guessing at the range of motion needed (like where the trigger pivots and how far it can be pressed).

Once we get the mesh properly rigged, I'll be happy to merge it!

atarng-magicleap:
Hello Toji, I updated the GLB file to include the transformations for the trigger and bumper button and have verified in the input profile link provided.

atarng-magicleap:
Hello @toji , I updated the GLB file to include the transformations for the trigger and bumper button and have verified in the input profile link provided.

toji:
Looks good, thank you for the update! I'm going to merge this now, but I do have two things to make note of that may or may not warrant a follow up patch:

 - The range of motion for the "squeeze" button seems to have it pushing much further into the controller than I expect the physical button does. This isn't a serious problem, it might just look a little funny to users if they push the button a little and the virtual version moves a lot.
 - The controller appears to sit lower on the Y axis and at more of a tilt than most other models (though it's somewhat similar to the `magicleap-one` controller). It's more common to see the controller aligned with the Z axis and with the center of the handle sitting at roughly the origin. This may not be a problem, depending on how the browser on the Magic Leap exposes the controller grip spaces. I have no way of testing that at the moment. In any case, the critical thing is that the controller model is positioned such that the physical and virtual controllers line up as closely as possible when the user is holding them. As long as that's the case it's fine.

Thank you again!

toji (approved):

nielsvanvelzen:
@toji the changes were merged into the `master` branch instead of `main`, was that intended?

toji:
Oh, no. Thanks for catching that, it should be main. I'll resolve that myself either today or tomorrow.

PR #265, opened 2024-09-03
Add support for Logitech MX Ink

cabanier:
First pass at support for the MX Ink.
The model still needs to be opened in blender to adjust for the rest position and add the buttons names and animations.


AdaRoseCannon:
Thank you for doing this so quickly!

It should probably be none.glb handed rather than right.glb

cabanier:
> Thank you for doing this so quickly!
> 
> It should probably be none.glb handed rather than right.glb

The pen is always surfaced as either left or right handed so I think this is correct.

AdaRoseCannon:
Could a user potentially have two, one in each hand?

cabanier:
> Could a user potentially have two, one in each hand?

I don't think that is supported but I will ask

cabanier:
Also, note that the user can still hold the regular controller in their other hand.

AdaRoseCannon:
I think it could still be "none" because it appears that the model is still the same between hands and the input profile. Is there some difference that means that we need to expose the handedness? It feels odd to me is all.

cabanier:
> I think it could still be "none" because it appears that the model is still the same between hands and the input profile. Is there some difference that means that we need to expose the handedness? It feels odd to me is all.

The name in this directory corresponds with the ["handedness"](https://immersive-web.github.io/webxr/#dom-xrinputsource-handedness) attribute of the inputSource. Since that will never be `none`, I don't think need to provide that file.

toji:
Ada is right: If it's the same mesh that's used in both hands you can upload it as 'none.glb' and specify in the JSON that it's used for either hand. It'll still come through with the appropriate handedness in-app. See the [HTC Vive asset](https://github.com/immersive-web/webxr-input-profiles/tree/main/packages/assets/profiles/htc-vive) for an example of how this works.

And yes, thanks for getting this up!

packages/registry/profiles/logitech/logitech-mx-ink.json
```diff
@@ -0,0 +1,31 @@
+{
+  "profileId": "logitech-mx-ink",
+  "fallbackProfileIds": [
+    "generic-trigger-squeeze-thumbstick"
```
toji, line 4: Is this the right fallback ID? It's hard for me to tell the exact capabilities from the marketing materials out there, but at first blush this seems more like it would fall back to `generic-trigger` or `generic-button`? Also, down in the components section you list "touch-pad", which seems to conflict with the "-thumbstick" in this profileID.

AdaRoseCannon:
This looks good to me!

AdaRoseCannon (approved):

cabanier:
@AdaRoseCannon @toji note that the model still needs to be optimized in blender to show the animations and the correct default position.

AdaRoseCannon:
I think it can't find the path:

image

I think you might need the override as described here: https://github.com/immersive-web/webxr-input-profiles/tree/601f7a2e0e6956ce431cf2e38c10b33e3e65eb34/packages/assets/tutorial#add-a-profilejson

ferminLR:
@cabanier I have prepared a model with the correct position and the buttons animations; but I am not sure which component  is which button, apart from the obvious like `tip-force`. I tried to figure out which one is which using a Quest3, but the profile I get when connecting the stylus is still `meta-quest-touch-plus`. Can you help me with this? Whenever I get to know the buttons ids, I'll make a PR with the model in no time.

@AdaRoseCannon the viewer can't find the path because by default it tries to load the left hand model; you just have to select `none` in the dropdown menu on the right.

cabanier:
> @cabanier I have prepared a model with the correct position and the buttons animations;

Thank you so much! that is awesome.

> but I am not sure which component is which button, apart from the obvious like `tip-force`. I tried to figure out which one is which using a Quest3, but the profile I get when connecting the stylus is still `meta-quest-touch-plus`. Can you help me with this? Whenever I get to know the buttons ids, I'll make a PR with the model in no time.

Unfortunately, you need v69 of the OS and Browser 35.0 to get proper support.
 I'd be happy to try it out and post a recording of your model.

gamepad 0 = front button (close to tip) only click - needs animation
gamepad 1 = back button (next to front button and touch area) only click - needs animation
gamepad 2 = unused
gamepad 3 = unused
gamepad 4 = tip (only force)
gamepad 5 = touch area between buttons
gamepad 6 = "docked" status.

ferminLR:
Thanks for your rapid response @cabanier! I just made the PR #266.

Just one thing, according to the profile:

```javascript
      "gamepad": {
        "mapping": "xr-standard",
        "buttons": [
          "xr-standard-trigger",
          "xr-standard-squeeze",
          null,
          null,
          "touch-pad",
          "tip-force",
          "docked"
        ],
        "axes": []
      }
``` 

The `touch-pad` is in gamepad 4 and `tip-force` in gamepad 5, instead of them being swapped as you said; I guess the correct way is the one from the profile and your list had a typo, so I made the model considering that.

Let me know if you find any errors and I'll update the model.

> Unfortunately, you need v69 of the OS and Browser 35.0 to get proper support.

I am on v69 PTC but can't update the Browser to 35.0. Not sure if there's a program to access beta versions of the browser.

cabanier:
> The `touch-pad` is in gamepad 4 and `tip-force` in gamepad 5, instead of them being swapped as you said; I guess the correct way is the one from the profile and your list had a typo, so I made the model considering that.

Yes, my list had a typo. Sorry about that!
> > Unfortunately, you need v69 of the OS and Browser 35.0 to get proper support.
> 
> I am on v69 PTC but can't update the Browser to 35.0. Not sure if there's a program to access beta versions of the browser.

I can send you a private build to try. Are you on the WebXR Discord? If so, you can dm me there at @rc

ferminLR:
> @AdaRoseCannon the viewer can't find the path because by default it tries to load the left hand model; you just have to select none in the dropdown menu on the right.

I was wrong, the model should appear no matter which handedness you choose in the viewer. The error appears because the left and right models aren't properly linked with the `node.glb` file. Fixed in #266.

PR #266, opened 2024-09-11
Update Logitech MX Ink model

ferminLR:
Added 3D model of the Logitech MX Ink stylus, with the correct position, orientation and the buttons animated.


ferminLR:
Tested with the latest Quest Browser version. Model position, orientation and button animation are correct (well, in fact, the button animations are a bit exaggerated to give a proper visual feedback, the buttons in the device barely move)

Also, when testing I noticed both left and right handed models weren't linked to the `none.glb` file. Now that is fixed.

toji:
Thank you! Checked out locally and verified that the buttons are indeed mapped as one would expect and that the mesh is showing up for left, right, and none handedness.

I'll leave @cabanier to verify that the pose of the pen matches the physical device when held, but otherwise LGTM to merge!

ferminLR:
I just noticed the viewer report a pair of errors for missing nodes (they don't correspond to any moving part in the controller), and the "Preparing a WebXR input profile mesh" tutorial recommends to create empty nodes in the mesh to remove those errors. I am gonna add those, just in case.

AdaRoseCannon:
I think adding the model override as described here, would let it work with none,left,and right handed inputs: https://github.com/immersive-web/webxr-input-profiles/tree/601f7a2e0e6956ce431cf2e38c10b33e3e65eb34/packages/assets/tutorial#add-a-profilejson

AdaRoseCannon:
Nevermind previous now deleted comment, I just saw you already included the override.

cabanier:
@ferminLR I pullled your model into my repo but I don't see any button animations.
I also see:

> Could not find xr_standard_trigger_pressed_min in the model
> Could not find xr_standard_squeeze_pressed_min in the model
> Could not find tip_force_pressed_min in the model
> Could not find docked_pressed_min in the model

cabanier:
> @ferminLR I pullled your model into my repo but I don't see any button animations. I also see:

I forgot to rebuilt. This is looking great! lgtm

cabanier (approved):

cabanier:
@toji Can you push this to the CDN when you have a chance?

toji:
Yeah, I'll see if I can get that out today.

cabanier:
> Yeah, I'll see if I can get that out today.

actually, can you wait? We might need some adjustment to the model.

PR #268, opened 2024-09-16
FIX #267: Mx Ink target ray alignment

### LINKED ISSUE #267: Logitech MX Ink: Target ray space not aligned with controller grip
The target ray is not correctly positioned and aligned in the `logitech-mx-ink` profile. The model matches the physical position and orientation of the device, and when comparing when the model drawn by the OS (by clicking in the menu button, so the OS windows appear) the models match but the rays don't. See attached images

Left: The model and its ray, properly aligned, as drawn by the OS shell.
Right: Inside WebXR (the profile viewer), the model is correctly posed, but the target ray isn't.

Some additional info:
- There is a transformation done to the controller mesh, because the target ray is just a line across the controller local Z axis, and the controller meshes axes, when viewed in Blender, don't match that direction.
- For any profile, the target ray viewed in the WebXR emulators in PC Chrome and Firefox, the Quest Browser, and Wolvic, are different; it seems the controller mesh transformation is browser dependant.
- AFAIK, There is no way to override this transformation of define the target ray space in the profile definition.
- The mesh could be edited to align the ray, of course, but that would make it not matching the device pose in real life.
- Other profiles suffer similar problems, e.g. `magicleap-one`

I understand this hasn't been a problem before because usually each platform only have had a small number of controllers compatible, so the browsers on those platforms had the model transformations setup for their controllers. Now that Meta is opening up the OS for third party controllers, **would it make sense to add an optional definition of the targetRaySpace in the profiles?**

ferminLR:
Partly fix of issue #267.

In a previous version of the Quest Browser, the target ray would not align with the stylus. @cabanier is taking care of the necessary browser changes. This PR updates the Logitech MX Ink model to properly align it with both the real device and the fixed target ray.


cabanier:
I think you need to update `left.glb` and `right.glb` as well

cabanier (commented):
I verified that the pen and ray line up correctly

cabanier:
@toji can you publish to the CDN?

toji:
Sorry that took me a few days, but it's published now! v1.0.16

cabanier:
Thanks @toji !

PR #272, opened 2025-03-12
Update Yvr touch v2 Controller Model

xkiti:
I found that in [this webxr application](https://threejs.org/examples/?q=ball#webxr_xr_ballshooter), the controller model does not fit the actual controller. like this:
image

So,I modified the initial rotation of the controller model.like this:
image


toji:
Thank you for testing the models and verifying the orientation! I appreciate the fix.

The PR seems like it needs a couple of adjustments before we can merge it, though:
 - It adds the .glb meshes to `packages/registry/profiles/yvr/` when they should actually be updates to the ones already in `packages/assets/profiles/yvr-touch-v2/`
 - This PR shows you adding `packages/assets/profiles/yvr-touch-v2/profile.json` and `packages/registry/profiles/yvr/yvr-touch-v2.json` as new files, but those already exist in the repository and appear to be identical to the ones in this PR, so I'm not sure why it's saying they need to be added? Perhaps try rebasing against `immersive-web:main`?

If we can get those issues sorted out, though, I'm happy to merge the fix! Thanks again!

xkiti:
Thank you for reminding. I made a small mistake, and now I have adjusted the catalog.

xkiti:
I made a small mistake, and now I am resubmitting a PR to fix issue #273.

PR #274, opened 2025-04-18
Add support for micro gestures

cabanier:
these are now supported on the oculus hand profile


toji (approved):
Thanks for the update!

Issue #269, opened 2024-11-12
Pico 4 Ultra Controllers


kalegd:
**Profile id**
N/A

**Describe the bug**
Users have reported to me that their Pico 4 Ultra controllers are not oriented correctly. I'm using the latest version of the input profiles, but the weird thing is that pico 4 ultra isn't even an option, but somehow their controller model is getting loaded? I'm a little confused on how to proceed and rectify this issue

What I know
* This repo only contains the following Pico controllers `pico-4`, `pico-g2`, `pico-neo2`, and `pico-neo3`
* The above is true when loaded from [jsdelivr](https://cdn.jsdelivr.net/npm/@webxr-input-profiles/assets@1.0/dist/profiles/profilesList.json) normally
* When the above link is loaded in the Pico browser, there are additional records for `pico-merline` and `pico-phoenix` (`pico-4' is removed)
* A request is then made for `oculus-touch-v2/profile.json`, but the profile-id comes in as `pico-sparrow`
* The url for the model says `oculus-touch-v2/left.glb`, but the model loaded is actually a pico 4 ultra controller

I'm guessing their browser does some weird overwriting thing?

**To Reproduce**
1. Use a Pico device or the Pico device emulator
2. Go to any webxr website that uses webxr-input-profiles like https://sliceofmusic.com
3. Begin an immersive XR session
4. Look at your controllers and see the incorrect orientation

**Expected behavior**
Generic fallback controller used, or the Pico 4 ultra controller is oriented correctly

**Screenshots**
Video curtesy of Akumelion from the Pico discord as I don't own this headset (I did verify this strange behavior with the emulator as well)

https://github.com/user-attachments/assets/af8b8755-0b76-4e9b-bdfc-a790fa77c4af

**3D engine (please complete the following information):**
 - Engine: three.js
 - Version r170

**Additional context**
Requesting advice on how to best proceed, if anyone has any contacts from the Pico browser/headset teams, or if there are any recommended alternative solutions I could pursue?

toji:
Hrm. That definitely sounds like they're intercepting the HTTP request and returning their own internal assets. Which I'm not exactly a fan of, but I guess could work out in some limited cases assuming that the assets that they deliver are actually well formatted and aligned. It's unfortunate that they don't seem to have taken the time to validate that here. (Also, this approach won't work for any sites not using that specific CDN.)

[...]

PR #275, opened 2025-06-09
Add PicoVR 4 Ultra controller models and profile

ziliang-bd:
Hi, This PR upload models that PICO 4U uses and it's expected to resolve the incorrect orientation pointed out by #269. Once it's been merged, we can start to work on that "overwriting issue" (also mentioned in #269). It's appreciated if anyone can help review it. ：-）


toji:
Checked it out locally and the rigging looks good! Thanks!

The only concern that I have is that the controller appears to be fairly low on the Y axis, at least compared to other assets in the repo. You can see the difference between this asset and the Meta Quest Plus controller in this overlay:

Controller-comparison

This may be intentional, as the device itself may be returning a grip space that is offset from the physical controller location by a bit and this is what's required to align it. Since I don't have a PicoVR headset I have no way to confirm that. I will note that if this kind of offset is required to align with the headset reported values then it's likely that any app-provided meshes that are intended to be held by the user (ie: a sword, gun, paintbrush, etc) will be mis-aligned on this device. A preferable solution would be to adjust the coordinates that the browser is returning to negate the offset so that the controller asset can be positioned at the origin. That would ensure the best compatibility with WebXR content.

Regardless, I'm happy to merge this is you can confirm that it is well aligned when rendered on the actual hardware.

ziliang-bd:
@toji Yes, this offset is used to solve the orientation issue. Thanks for pointing out!  

And I agree that it's more appropriate to adjust coordinates in Browser so I upload a newest one and it centers at (0, 0) and faces one axis.

We will adjust this mis-align issue in browser. When we release that version,  any app-provided meshes designed by the user will be  aligned correctly on this device. (so as this model) :-)  Thanks!!

ziliang-bd:
Thanks for providing this context, it's helpful to know this. I submit a new one with handles at the origin, pointing down the Z axis. And I will leave a notice about this to make sure our designers provide well-fit models in the future.

toji:
This looks great now, thank you! I really appreciate the effort. I'll merge it once it passes the checks and then get a new version deployed to NPM.

ziliang-bd:
Thank you :-). And the unexpected orientation of model of PICO 4, I will adjust the coordinates in the browser first to negate the offset, required by this new pico-4u model, and then submit a new PR for that model.

PR #280, opened 2025-11-13
Expose oculus-hand profile by adding assets

felixtrz:
# Add assets for the `oculus-hand` profile to enable CDN distribution and developer usage.

The `oculus-hand` profile has existed in the registry (`packages/registry/profiles/oculus/oculus-hand.json`) but was never exposed through the assets package. Without corresponding assets, the profile cannot be built into the distribution and remains unavailable on the CDN, making it effectively unusable by developers (it is not available from https://cdn.jsdelivr.net/npm/@webxr-input-profiles/assets@1.0.20/dist/profiles/profilesList.json).

## Changes
- Added `packages/assets/profiles/oculus-hand/profile.json` configuration
- Added `left.glb` and `right.glb` hand models (reused from `generic-hand`)
- The profile will now be included in the built `profilesList.json` and distributed via CDN

## Why reuse generic-hand models?
The `oculus-hand` profile is designed for hand tracking (not physical controllers), and the existing `generic-hand` models provide appropriate visual representation. The profile defines hand gesture inputs including:
- Pinch select (`xr-standard-trigger`)
- Swipe gestures (left, right, forward, backward)
- Tap-thumb gesture
- Menu button (left hand only)

## Impact
Developers working with Oculus/Meta hand tracking can now load and visualize the `oculus-hand` profile through the standard WebXR input profiles library and CDN.


felixtrz:
@cabanier I see, this may be the case when we added this profile, but now it does add new information - it has the gestures in the gamepad buttons array which the generic-hand doesn't have; so by skipping this, developers will never have access to the gesture button mapping.

cabanier:
Can you make it so that you grab the profiles from `generic-hand`?
That way you are pointing to the same assets and you won't have to add new glb's

simonedevit:
Interested in this topic as I submitted a similar PR https://github.com/immersive-web/webxr-input-profiles/pull/278 two months ago.

There are two practical considerations that make it difficult to reuse [generic-hand/left.glb](https://github.com/immersive-web/webxr-input-profiles/blob/main/packages/assets/profiles/generic-hand/left.glb) and [generic-hand/right.glb](https://github.com/immersive-web/webxr-input-profiles/blob/main/packages/assets/profiles/generic-hand/right.glb) directly for the `oculus-hand` profile:


**1. Cross-folder asset references are not allowed by the schema**

Anything like:
```json
{
    "profileId" : "oculus-hand",
    "overrides" : {
        "left": {
            "assetPath": "profiles/generic-hand/left.glb"
        }
    }
}
```
is invalid because it violates the filename constraint `^[a-z0-9]+(-[a-z0-9]+)*\.glb$` defined in the [layout schema](https://github.com/immersive-web/webxr-input-profiles/blob/main/packages/assets/schemas/layout.schema.json#L14).

`assetPath` must be a plain filename, without directories or relative paths. This prevents any direct reuse of assets across folders.

**2. Even if the schema restriction were bypassed, `generic-hand` lacks the required components**

Even assuming we introduce workarounds such as modifying the validator, manually rewriting the `assetPath` at runtime, or injecting a fallback before creating the `MotionController` (all of which I would consider bad practices), the `generic-hand` models still do not define the components needed for `oculus-hand`, such as `menu_pressed_min`, `swipe_left_pressed_min`, `swipe_right_pressed_min`, and others.

This results in errors like the following:

image
</br></br>
Given these constraints, it might be worth considering a straightforward duplication of the  `generic-hand` meshes within the `oculus-hand` folder, along with the additional component nodes required by this profile. This is the approach I followed in my earlier PR, and it avoids schema issues, removes the need for workarounds, and keeps the profile self-contained, though of course there may be alternative solutions the maintainers prefer.

cabanier:
@felixtrz what do you think of the comments of @simonedevit? Could you apply these to your PR?

cabanier (approved):

PR #281, opened 2026-05-13
DON'T MERGE!! Label nodes for oculus-touch-v3

AdaRoseCannon:
Response to https://github.com/immersive-web/webxr/issues/1428

I took a stab at seeing what it would look like for a single controller.

This is just a POC to see if this is the kind of thing that we had in mind.

Adds`<rootNodeName>_label` nodes to left/right GLBs at positions chosen to face outward from the controller surface, so experiences can anchor tooltips/text to each button, trigger, squeeze, thumbstick, and thumbrest.

- GLBs: embed `_label` empty nodes (local +Y outward) under each component's root subtree.
- profile.json: declare matching `labelTransform` overrides (position
  + quaternion in gltf.scene-local / grip space) so consumers who can't parse the GLB still get the placement.
- layout.schema.json: allow `labelTransform` on a component.

/facetoface


AdaRoseCannon:
A visual:
image

AdaRoseCannon:
@toji I've made a little tool so I can easily annotate every model and JSON, can you review this to make sure it is what you had in mind and if it is then I will do every single one.

